PR #79, opened 2026-03-04
Auto-inject version from package.json

IanLindsley:
## Summary
- Make `package.json` the single source of truth for the worker version
- A prebuild script (`scripts/generate-version.mjs`) generates `src/generated/version.ts` with an `APP_VERSION` constant
- Health endpoint now imports `APP_VERSION` instead of hardcoding `'2.0.0'`
- Prompt overrides support `{{version}}` template variables that get replaced at runtime after resolution
- npm lifecycle hooks (`prebuild`, `predev`, `version`) keep the generated file in sync automatically

## How to update the version after this change
1. Run `pnpm version patch` (or `minor`/`major`)
2. The `version` lifecycle hook regenerates the constant and stages it
3. Commit, push, deploy — the new version flows everywhere automatically

## Test plan
- [x] `pnpm check` — TypeScript compiles
- [x] `pnpm lint` — no new warnings
- [x] `pnpm test` — 291 tests pass (10 new template tests)
- [x] `pnpm build` — prebuild hook runs, worker builds
- [x] `pnpm architecture` — no circular dependencies
- [ ] CI passes
- [ ] Verify `/health` endpoint shows version from `package.json` on staging


package.json
```diff
@@ -4,9 +4,13 @@
...
   "type": "module",
   "scripts": {
+    "generate:version": "node scripts/generate-version.mjs",
+    "predev": "node scripts/generate-version.mjs",
     "dev": "wrangler dev",
     "deploy": "wrangler deploy",
+    "prebuild": "node scripts/generate-version.mjs",
     "build": "wrangler build",
+    "version": "node scripts/generate-version.mjs && git add src/generated/version.ts",
     "test": "vitest run",
```
IanLindsley: **[MEDIUM]** CI fragility -- test/check/lint/architecture jobs depend on checked-in generated file

Only `pnpm build` triggers the `prebuild` hook that generates `src/generated/version.ts`. The CI `test`, `check`, `lint`, and `architecture` jobs work solely because this file is checked into git.

If someone later adds `src/generated/` to `.gitignore` (the natural instinct for generated files) or deletes the file, four CI jobs break with cryptic import errors.

**Recommendation:** Add a `pretest` lifecycle hook:

```json
"pretest": "node scripts/generate-version.mjs",
```

This makes the test job self-sufficient at negligible cost (a few milliseconds). You could also consider adding `precheck` and `prelint` but those are less standard -- the `pretest` hook alone covers the most common CI failure mode.

src/utils/template.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Supports {{name}} placeholders in prompt text. Unknown variables are left as-is.
+ */
+
+import { APP_VERSION } from '../generated/version.js';
+import { PromptSlot, PROMPT_OVERRIDE_SLOTS } from '../types/prompt-overrides.js';
+
+/** Built-in template variables available in all prompt slots. */
+const TEMPLATE_VARIABLES: Record<string, string> = {
+  version: APP_VERSION,
```
IanLindsley: **[LOW]** TEMPLATE_VARIABLES is evaluated once at module load

This captures `APP_VERSION` at import time. That is correct for Cloudflare Workers (modules re-evaluate per isolate), but worth a brief inline comment noting the static binding is intentional. Future maintainers might try to add request-scoped values (e.g., user org, environment) to this map and be surprised when they do not update per-request.

Suggestion:
```ts
/** Built-in template variables available in all prompt slots.
 * Evaluated once at module load — do not add request-scoped values here.
 * Pass a custom variables map to replaceTemplateVariables() instead.
 */
```

src/index.ts
```diff
@@ -38,7 +39,7 @@ export { UserQueue, UserSession };
 const app = new Hono<{ Bindings: Env }>();
 
 // Health check - no auth required
-app.get('/health', (c) => c.json({ status: 'healthy', version: '2.0.0' }));
+app.get('/health', (c) => c.json({ status: 'healthy', version: APP_VERSION }));
```
IanLindsley: **[LOW]** Admin GET prompt-overrides endpoint does not apply template variables

The `/health` endpoint now correctly shows the resolved version. However, the admin `GET /api/v1/admin/orgs/:org/prompt-overrides` endpoint (line 350) returns `resolved` from `resolvePromptOverrides()` without calling `applyTemplateVariables()`. This means admins see raw `{{version}}` in the resolved preview rather than the substituted value.

This may be intentional (admins want to see template source), but it is inconsistent with what users actually receive. Consider:
1. Adding a `resolved_with_templates` field alongside `resolved`, or
2. Adding a comment documenting this as intentional behavior.

PR #86, opened 2026-03-11
Add request_audio tool and fix webhook audio forwarding

IanLindsley:
## Summary

- **Bug fix**: Webhook callbacks (`onComplete`) now forward `voice_audio_base64` from `ChatResponse` to the gateway. Previously, TTS audio was generated but silently dropped — the WhatsApp gateway never received it.
- **New feature**: Added `request_audio` built-in tool so Claude can flag text responses for TTS when users ask to "hear", "listen to", or "read aloud" content. Previously, TTS only triggered for `message_type === 'audio'` (voice messages).
- **System prompt guidance**: Added audio response section instructing Claude when to invoke `request_audio`.

## Files changed

| File | Change |
|------|--------|
| `src/services/progress/callback.ts` | Forward `voice_audio_base64` in complete callback |
| `src/services/audio/types.ts` | Add `AudioContext` interface |
| `src/services/audio/index.ts` | Export `AudioContext` |
| `src/services/claude/tools.ts` | Add `buildRequestAudioTool()`, update `isBuiltInTool`, update `buildAllTools` |
| `src/services/claude/orchestrator.ts` | Wire `audioContext` through options/context, add `handleRequestAudio()` |
| `src/durable-objects/user-session.ts` | Add `buildAudioContext()`, `maybeGenerateAudio()`, update TTS trigger |
| `src/services/claude/system-prompt.ts` | Add audio tool guidance section |
| `tests/unit/progress-callback.test.ts` | 5 new tests for audio forwarding |
| `tests/unit/claude-tools.test.ts` | Update counts, add `request_audio` tests |
| `tests/unit/audio-service.test.ts` | 3 new `AudioContext` tests |
| `tests/unit/system-prompt.test.ts` | 2 new audio guidance tests |

## Test plan

- [x] `pnpm check` — TypeScript compiles
- [x] `pnpm lint` — No lint errors
- [x] `pnpm test` — All 313 tests pass (15 test files)
- [ ] CI passes
- [ ] Manual test: send text "I want to listen to John 1:1" → verify `voice_audio_base64` is non-null in response

### DIFF
--- a/src/services/claude/orchestrator.ts
+++ b/src/services/claude/orchestrator.ts
@@ -572,8 +572,11 @@ async function handleUpdateMemory(input: unknown, ctx: OrchestrationContext): Pr
   return result;
 }
 
-function handleRequestAudio(ctx: OrchestrationContext): string {
-  ctx.audioContext?.requestAudio();
+function handleRequestAudio(ctx: OrchestrationContext): unknown {
+  if (!ctx.audioContext) {
+    return { error: 'Audio responses are not available for this session.' };
+  }
+  ctx.audioContext.requestAudio();
   return 'Audio response requested. Your text response will be converted to speech.';
 }
 
--- a/src/services/claude/tools.ts
+++ b/src/services/claude/tools.ts
@@ -207,7 +207,7 @@ export function buildAllTools(
     buildGetToolDefinitionsTool(),
     buildReadMemoryTool(),
     buildUpdateMemoryTool(),
-    buildRequestAudioTool(),
+    buildRequestAudioTool(), // Always available — TTS is a platform capability, not org-gated
   ];
 
   if (opts?.hasModes) {
--- a/src/services/progress/callback.ts
+++ b/src/services/progress/callback.ts
@@ -55,7 +55,7 @@ export class ProgressCallbackSender {
   async sendComplete(text: string, voiceAudioBase64?: string | null): Promise<void> {
     await this.post({
       type: 'complete',
-      text,
+      ...(text ? { text } : {}),
       ...(voiceAudioBase64 ? { voice_audio_base64: voiceAudioBase64 } : {}),
     });
   }
--- a/src/durable-objects/user-session.ts
+++ b/src/durable-objects/user-session.ts
@@ -607,8 +607,7 @@ export class UserSession {
     const { memoryStore, formattedTOC } = await this.loadMemoryContext(logger);
     const modeContext = this.buildModeContext(orgModes, activeModeName);
     const audioContext = this.buildAudioContext();
-    const startTime = Date.now();
-    const responses = await orchestrate(messageText, {
+    const responses = await this.runOrchestration(messageText, {
       env: this.env,
       catalog,
       history,
@@ -625,10 +624,6 @@ export class UserSession {
       clientId: body.client_id,
       logger,
       callbacks,
-    });
-    logger.log('phase_orchestration_complete', {
-      response_count: responses.length,
-      duration_ms: Date.now() - startTime,
     });
     const voiceAudioBase64 = await this.maybeGenerateAudio(
       body,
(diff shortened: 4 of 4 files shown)


src/services/claude/orchestrator.ts
```diff
@@ -565,6 +572,11 @@ async function handleUpdateMemory(input: unknown, ctx: OrchestrationContext): Pr
   return result;
 }
 
+function handleRequestAudio(ctx: OrchestrationContext): string {
+  ctx.audioContext?.requestAudio();
+  return 'Audio response requested. Your text response will be converted to speech.';
```
IanLindsley: **[MEDIUM]** `handleRequestAudio` lies to Claude when `audioContext` is undefined

When `audioContext` is `undefined` (it is optional on `OrchestratorOptions`), the optional chain `ctx.audioContext?.requestAudio()` silently does nothing, but the function still returns `'Audio response requested. Your text response will be converted to speech.'`. Claude believes audio will be generated, but it won't be -- the `audioRequested` flag was never set, so `maybeGenerateAudio` will skip TTS.

Today this is safe because `processChat` always provides `audioContext`. But if any future caller of `orchestrate()` omits it, the tool silently misbehaves.

**Recommendation:** Return an error message when `audioContext` is undefined, similar to how `handleReadMemory` returns `{ error: 'Memory is not available...' }` when `memoryStore` is undefined:

```typescript
function handleRequestAudio(ctx: OrchestrationContext): string | { error: string } {
  if (!ctx.audioContext) {
    return { error: 'Audio responses are not available for this session.' };
  }
  ctx.audioContext.requestAudio();
  return 'Audio response requested. Your text response will be converted to speech.';
}
```

src/services/progress/callback.ts
```diff
@@ -252,8 +257,8 @@ export function createWebhookCallbacks(
       incrementalSender?.complete();
       const fullText = response.responses.join('\n');
       const delta = fullText.slice(lastSentText.length);
-      if (delta) {
-        void sender.sendComplete(delta);
+      if (delta || response.voice_audio_base64) {
+        void sender.sendComplete(delta, response.voice_audio_base64);
```
IanLindsley: **[MEDIUM]** `sendComplete` sends empty `text` when only audio is present

When all text was already sent via incremental progress (so `delta` is `''`) but `voice_audio_base64` is present, this calls `sender.sendComplete('', response.voice_audio_base64)`. The gateway receives a `complete` payload with `text: ''` and `voice_audio_base64: '...'`.

Is this the intended contract? The WhatsApp gateway (or other consumers) may interpret an empty `text` field as "no response" and skip rendering. Alternatively, some gateways may not handle `text: ''` gracefully.

**Recommendation:** Either:
1. Document that `text` can be empty when only audio is present (if the gateway handles it), or
2. Omit the `text` field when it's empty, similar to how `voice_audio_base64` is omitted when null:

```typescript
void sender.sendComplete(delta || undefined, response.voice_audio_base64);
```

And update `sendComplete` to handle `text?: string`.

src/services/claude/tools.ts
```diff
@@ -191,6 +207,7 @@ export function buildAllTools(
     buildGetToolDefinitionsTool(),
     buildReadMemoryTool(),
     buildUpdateMemoryTool(),
+    buildRequestAudioTool(),
```
IanLindsley: **[LOW]** `request_audio` unconditionally included -- add a clarifying comment

Unlike `list_modes`/`switch_mode` which are gated behind `hasModes`, `request_audio` is always exposed. This appears intentional (TTS is a platform capability available to all orgs via the `AI` binding), but it breaks the established pattern where feature-specific tools are conditionally included.

A brief comment explaining *why* this is unconditional would help future reviewers not flag it:

```typescript
// request_audio is always available -- TTS is a platform capability, not org-configured
buildRequestAudioTool(),
```

PR #93, opened 2026-03-11
Merge mode overrides on PUT instead of replacing

IanLindsley:
## Summary
- The PUT `/modes/:modeName` endpoint previously replaced the entire mode object on update, meaning callers that only sent a subset of overrides would silently wipe the rest
- This was the root cause of `translation-coach` losing 4 of its 6 override slots — Baruch's `create_or_update_mode` tool only sent `methodology` and `instructions`, wiping `identity`, `closing`, `tool_guidance`, `memory_instructions`, and `client_instructions`
- Now uses the existing `mergePromptOverrides()` function to merge incoming overrides with existing ones (send `null` to explicitly remove a slot)
- Adds detailed before/after logging of override state for audit and data recovery

## Test plan
- [x] All 318 unit tests pass
- [x] TypeScript type check passes
- [x] ESLint passes
- [x] Architecture check passes
- [ ] CI passes on Linux (includes e2e tests)
- [ ] Verify partial mode updates preserve existing overrides
- [ ] Verify sending `null` for a slot removes it

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -603,12 +603,11 @@ function logAdminAction(action: string, org: string, details: Record<string, unk
   );
 }
 
-/** Upsert a mode into an OrgModes array, merging overrides with any existing mode. */
 /**
  * Upsert a mode into an OrgModes array, merging overrides with any existing mode.
  * Mutates orgModes.modes in-place (splice/push) and returns the result.
  */
-function upsertMode(
+export function upsertMode(
   orgModes: OrgModes,
   modeInput: PromptMode,
   org: string
--- a/tests/unit/upsert-mode.test.ts
+++ b/tests/unit/upsert-mode.test.ts
@@ -1,0 +1,90 @@
+import { describe, it, expect } from 'vitest';
+import { upsertMode } from '../../src/index.js';
+import { OrgModes, MAX_MODES_PER_ORG } from '../../src/types/prompt-overrides.js';
+
+function makeOrgModes(...modes: OrgModes['modes']): OrgModes {
+  return { modes };
+}
+
+describe('upsertMode - new mode creation', () => {
+  it('creates a new mode when none exists', () => {
+    const orgModes = makeOrgModes();
+    const result = upsertMode(orgModes, { name: 'test', overrides: { identity: 'Hello' } }, 'o');
+    expect(result.ok).toBe(true);
+    if (result.ok) expect(result.savedMode.overrides.identity).toBe('Hello');
+    expect(orgModes.modes).toHaveLength(1);
+  });
+
+  it('returns error when MAX_MODES_PER_ORG exceeded', () => {
+    const modes = Array.from({ length: MAX_MODES_PER_ORG }, (_, i) => ({
+      name: `mode-${i}`,
+      overrides: {},
+    }));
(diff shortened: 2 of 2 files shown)


src/index.ts
```diff
@@ -606,6 +603,44 @@ function logAdminAction(action: string, org: string, details: Record<string, unk
...
+    const existing = orgModes.modes[existingIdx]!;
+    logAdminAction('upsert_mode_before', org, {
+      mode_name: modeInput.name,
+      existing_overrides: existing.overrides,
+      incoming_overrides: modeInput.overrides,
+    });
+    const savedMode: PromptMode = {
+      ...existing,
+      ...modeInput,
+      overrides: mergePromptOverrides(existing.overrides, modeInput.overrides),
```
IanLindsley: **[MEDIUM]** Spread `...modeInput` silently wipes `label` and `description` on partial updates

The `...existing, ...modeInput` spread means that if the caller sends `{ overrides: { methodology: "new" } }` without `label` or `description`, those fields become `undefined` on the saved mode -- the same "silent wipe" problem this PR fixes for overrides.

For example, if the existing mode has `label: "Translation Coach"` and the caller only sends `methodology`, the saved mode will have `label: undefined`.

**Recommendation:** Only spread defined scalar fields from `modeInput`, or apply the same "merge if present, skip if absent" pattern used for overrides:

```ts
const savedMode: PromptMode = {
  ...existing,
  // Only overwrite scalars that were explicitly provided
  ...(modeInput.label !== undefined && { label: modeInput.label }),
  ...(modeInput.description !== undefined && { description: modeInput.description }),
  overrides: mergePromptOverrides(existing.overrides, modeInput.overrides),
};
```

IanLindsley (commented):
## Follow-up Review (3rd pass)

### Resolved

1. **[MEDIUM]** `...modeInput` spread silently wipes `label`/`description` -- **Fixed** (commit 2). Conditional spread with `??` fallback at `src/index.ts:626-629` correctly preserves existing values when the caller omits them.

2. **[MEDIUM]** Return type requires `savedMode` even on error path -- **Fixed** (commit 2). Clean discriminated union `{ ok: true; savedMode } | { ok: false; error }` at `src/index.ts:614`. Caller narrows correctly with `if (!result.ok)`.

3. **[LOW]** Mutation side-effect without documentation -- **Fixed** (commit 2). JSDoc at `src/index.ts:606-609` clearly states in-place mutation.

4. **[MEDIUM]** No unit tests for `upsertMode` -- **Fixed** (commit 3). Eight tests across three `describe` blocks covering:
   - New mode creation and capacity limit rejection
   - Override merging (preserves untouched slots), `null`-to-remove, update at max capacity
   - Scalar field preservation (label and description kept when omitted) and overwrite (label replaced when provided)

5. **[LOW]** Audit logging sensitivity -- Acknowledged as acceptable; no change needed.

6. **[LOW]** Duplicate JSDoc comment -- **Fixed** (commit 3). Single JSDoc block remains at `src/index.ts:606-609`.

### New Issues

None found.

**Status:** 6/6 issues resolved. All previously flagged findings have been addressed.

**Recommendation:** APPROVE -- the PR is clean. The merge logic is correct, well-tested, and the discriminated union return type makes the success/error paths unambiguous.

PR #98, opened 2026-03-12
fix: eliminate silent error swallowing across codebase

IanLindsley:
## Summary

- **Eliminates 43+ instances** of silent error handling (empty catches, `console.error`, `void` fire-and-forget, unlogged 500 responses, broken logger correlation) across 6 source files
- **Adds `safeAsync()` and `withEndpointLogging()` helpers** to `logger.ts` for consistent async error capture and endpoint entry/exit logging
- **Replaces all `console.error`/`console.warn`** with structured, request-correlated `RequestLogger` calls so every failure is observable in production logs
- **Adds 10s fetch timeout** to webhook callbacks to prevent DoS from unresponsive endpoints
- **Migrates admin logging** from ad-hoc `logAdminAction` function to inline structured logging with request IDs

### Files changed

| File | What changed |
|------|-------------|
| `src/utils/logger.ts` | Added `safeAsync()`, `withEndpointLogging()` |
| `src/services/progress/callback.ts` | Structured logger, fetch timeout, replaced void calls |
| `src/durable-objects/user-queue.ts` | Fixed silent SSE catches, endpoint logging, alarm recovery |
| `src/durable-objects/user-session.ts` | Fixed unlogged 500s, logger context loss, endpoint logging |
| `src/index.ts` | Migrated admin logging, added queue proxy logging |
| `src/services/claude/orchestrator.ts` | Added Claude stream error listener |
| `tests/unit/progress-callback.test.ts` | Updated assertions for AbortController signal |

## Test plan

- [x] `pnpm check` — TypeScript compilation clean
- [x] `pnpm lint` — 0 errors (4 pre-existing warnings only)
- [x] `pnpm test` — 337/337 tests pass
- [x] `pnpm architecture` — No circular dependencies
- [x] Grep audit: `console.*` only in `logger.ts`, zero `void` fire-and-forget, zero empty catches in changed files

### DIFF
--- a/src/durable-objects/user-session.ts
+++ b/src/durable-objects/user-session.ts
@@ -98,6 +98,11 @@ function createErrorResponse(
   return Response.json({ error, code, message }, { status });
 }
 
+function storageErrorResponse(err: unknown): Response {
+  const msg = err instanceof Error ? err.message : String(err);
+  return createErrorResponse('Storage error', 'INTERNAL_ERROR', msg, 500);
+}
+
 export class UserSession {
   private state: DurableObjectState;
   private env: Env;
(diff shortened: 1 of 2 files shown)


IanLindsley (commented):
## Follow-up Review (Re-review #2)

### Previously Flagged Issues

#### Resolved

- **[MEDIUM] `runSafe` silently swallows errors when no logger is provided** (callback.ts:274) -- **Confirmed fixed.** `runSafe` now requires `RequestLogger` (not optional) and delegates directly to `safeAsync(logger, event, fn)`. The `fn().catch(() => {})` fallback is gone. `IncrementalProgressSender` and `createWebhookCallbacks` both require `RequestLogger` at the type level, enforcing correctness at compile time.

- **[LOW] Double-wrapping of `withEndpointLogging` and inner try-catch in DO handlers** (user-session.ts) -- **Confirmed fixed.** `withEndpointLogging` now accepts an optional `onError` callback parameter. Six DO endpoints (`handleGetPromptOverrides`, `handleUpdatePromptOverrides`, `handleDeleteHistory`, `handleDeletePromptOverrides`, `handleGetMemory`, `handleDeleteMemory`) use `storageErrorResponse` as the `onError` callback, eliminating the redundant inner try-catch blocks. The `withEndpointLogging` implementation correctly calls `onError(err)` in the rejection handler before falling through to re-throw, which is the right behavior.

### New Issues

None. The fixes are clean and surgical. The `onError` callback addition to `withEndpointLogging` is a well-designed extension -- it handles the common pattern of "log the error AND return a structured error response" without requiring double-wrapping.

**Status:** 2/2 issues resolved. **Recommendation:** APPROVE

PR #118, opened 2026-04-13
feat: split /api/v1/chat into explicit transport endpoints

IanLindsley:
## Summary

Adds `POST /api/v1/chat/stream` and `POST /api/v1/chat/callback` as explicit-transport siblings of `/api/v1/chat`. Each new endpoint locks its transport mode rather than inferring it from the request body, so gateway consumers no longer need to know that passing `progress_callback_url` silently flips `/chat` from SSE to webhook mode.

- `POST /api/v1/chat/stream` — always SSE. Rejects `progress_callback_url`, `progress_mode`, `progress_throttle_seconds` with `400`.
- `POST /api/v1/chat/callback` — always `202` + webhook. Requires `progress_callback_url` and `message_key`; `400` if either is missing.
- `POST /api/v1/chat` — **unchanged** in this PR. Keeps legacy body-dispatch behavior. Will become final-only JSON in v2.14.0 once consumers migrate.

## Context

Denis's Telegram gateway (`bt-servant-telegram-gateway`) was receiving SSE `status` / `progress` frames leaking into user chats because `/chat`'s implicit dispatch returns a stream whenever `progress_callback_url` is absent. Making transport explicit lets the Telegram gateway consume a clean final-only JSON contract once `/chat` flips in v2.14.0.

Per Telegram exchange 2026-04-13, target contract accepted by Denis:
- `/api/v1/chat` → final-only JSON
- `/api/v1/chat/stream` → SSE
- `/api/v1/chat/callback` → `202` + webhook

## Rollout plan

This PR is **additive and non-breaking**. Follow-up PRs:

| PR | Repo | Effect |
|---|---|---|
| 1 (this PR) | bt-servant-worker | Add `/chat/stream` + `/chat/callback`. `/chat` unchanged. v2.13.0. |
| 2 | bt-servant-web-client | Migrate from `/chat` (SSE) to `/chat/stream`. |
| 3 | bt-servant-whatsapp-gateway | Migrate from `/chat` + `progress_callback_url` to `/chat/callback`. |
| 4 | bt-servant-worker | Flip `/chat` to final-only JSON. v2.14.0. |

Tracking issues for PR 2 and PR 3 will be filed in the sibling repos immediately after this PR opens.

## Changes

- `src/index.ts` — three explicit Hono routes; `handleChatRequest` / `buildDOChatRequest` / `validateChatBody` all thread a `ChatTransport` discriminator.
- `src/durable-objects/user-do.ts` — DO fetch handler dispatches `/chat`, `/chat/stream`, `/chat/callback` into `handleUnifiedChat(request, transport)`.
- `src/types/engine.ts` — new `ChatTransport = 'legacy' | 'stream' | 'callback'` type.
- `README.md` — new Chat section with all three endpoints and a v2.14.0 flip note.
- `tests/unit/chat-transport-dispatch.test.ts` — 12 pure-function tests for `validateChatBody` per-transport rules.
- `tests/e2e/chat-transport-dispatch.test.ts` — 3 DO path-dispatch smoke tests (skipped on Windows, run in CI on Linux).
- Version bump to **2.13.0**.

## Test plan

- [x] `pnpm test` — 388/388 pass locally (Windows; e2e tests skipped locally and run in CI on Linux)
- [x] `pnpm check` — clean
- [x] `pnpm lint` — 0 errors (5 pre-existing warnings unrelated to this PR)
- [x] `pnpm architecture` — 0 violations
- [ ] CI green on branch
- [ ] Manual verification against `wrangler dev` — `/chat/stream` returns SSE frames, `/chat/callback` returns 202 + webhook delivery, `/chat` unchanged, negative-tests 400 on forbidden fields

## Notes for reviewers

- `InternalQueueEntry` and `processQueueEntry` intentionally still route the queue branch on `body.progress_callback_url` presence. This remains correct because the new worker-side validation guarantees `/chat/stream` bodies never have the field and `/chat/callback` bodies always do — so the existing body-based discriminator is a perfect proxy for the requested transport. No queue schema change was needed in this PR.
- Denis's Telegram gateway is still blocked on SSE leaking into user chats until PR 4 lands. Priority should be on getting PR 2 and PR 3 through the pipeline so PR 4 can ship soon.


IanLindsley:
Tracking issues filed for the follow-up consumer migrations (PR 2 and PR 3 in the rollout plan above):

- bt-servant-web-client: https://github.com/unfoldingWord/bt-servant-web-client/issues/24 (migrate to `/chat/stream`)
- bt-servant-whatsapp-gateway: https://github.com/unfoldingWord/bt-servant-whatsapp-gateway/issues/18 (migrate to `/chat/callback`)

Both must land and deploy before the v2.14.0 flip PR on this repo.

IanLindsley (commented):
Inline comments only — see overall summary in PR review comment.

src/durable-objects/user-do.ts
```diff
@@ -214,7 +230,14 @@ export class UserDO {
...
 
-    const isCallbackMode = !!body.progress_callback_url;
+    // Resolve the effective delivery mode for this request:
+    //   - stream   → always SSE
+    //   - callback → always webhook
+    //   - legacy   → pick based on body.progress_callback_url presence
+    // The effective delivery is what matters for queue semantics and for
+    // which background runner we invoke.
+    const isCallbackDelivery =
+      transport === 'callback' || (transport === 'legacy' && !!body.progress_callback_url);
```
IanLindsley, line 281: **[MEDIUM]** Fragile invariant: `processQueueEntry` still re-derives delivery from `body.progress_callback_url`, which silently desyncs from `transport` if the worker validation is ever bypassed.

**Why:** The PR description correctly observes that today the worker's `validateTransportFields` is the only thing keeping `body.progress_callback_url` in lockstep with the transport route. Under that invariant the four cases listed in the PR description all hold. But the DO and the queue have no defense-in-depth: if a future refactor adds a new chat route, a test that calls `stub.fetch('/chat/callback', ...)` directly, or simply a bug that drops the worker validation, the failure mode is silent and surprising:

- `/chat/callback` request arrives at the DO with no `progress_callback_url` in the body.
- `handleUnifiedChat` correctly sets `isCallbackDelivery = true` (because `transport === 'callback'`) and returns `202` to the client.
- In the immediate path, `buildWebhookCallbacks(body, ...)` returns `undefined` because `!body.progress_callback_url`. `processChat` runs, completes, and `await callbacks?.onComplete?.(response)` is a silent no-op. The user gets `202` and never hears back.
- In the queued path, `processQueueEntry` recomputes `isCallbackMode = !!body.progress_callback_url = false` and routes the entry to `processSSEEntry`. But there is no SSE writer registered (because `enqueueAndReturn` returned 202, not a stream), so the `sendEvent` calls all fall through `clientDisconnected` checks and do nothing. Same silent swallow.

Neither mode logs a clear "transport mismatch" error. The CLAUDE.md "never silently swallow errors" rule applies here even though no `catch` is involved — the failure mode is invisible by construction.

**Suggestion:** Pick one of the following (in increasing order of robustness):

1. **Cheap fix — DO-side defense in depth.** Mirror the worker validation right after `body = await request.json()` in `handleUnifiedChat`:
   ```ts
   if (transport === 'callback' && (!body.progress_callback_url || !body.message_key)) {
     logger.error('do_transport_invariant_violated', new Error('callback transport missing required fields'), {
       transport,
       has_progress_url: !!body.progress_callback_url,
       has_message_key: !!body.message_key,
     });
     return Response.json({ error: 'progress_callback_url and message_key are required on /chat/callback' }, { status: 400 });
   }
   if (transport === 'stream' && body.progress_callback_url) {
     logger.error('do_transport_invariant_violated', new Error('stream transport with progress_callback_url'), { transport });
     return Response.json({ error: 'progress_callback_url is not valid on /chat/stream' }, { status: 400 });
   }
   ```
   Cheap, makes the invariant explicit, and surfaces the bug loudly if the worker layer ever drops validation.

2. **Robust fix — make transport explicit on the queue entry.** Add `transport: ChatTransport` to `InternalQueueEntry`, populate it in `enqueueAndReturn`, and have `processQueueEntry` dispatch on `entry.transport` instead of `body.progress_callback_url`. This eliminates the proxy entirely and the queue logic stops depending on an external invariant. (Existing in-flight queue entries from previous deploys would lack `transport`; fall back to the body-URL check for those.)

Fix 1 is enough to unblock this PR — it adds about ten lines and zero behavior change in the happy path. Fix 2 is the right long-term answer and could be a follow-up. Either way, please don't ship the current state where the only thing keeping the queue path correct is a comment in the PR description.

src/index.ts
```diff
@@ -900,49 +929,92 @@ function validateLanguageHint(hint: unknown): string | null {
...
+      return 'progress_callback_url is required on /api/v1/chat/callback';
+    }
+    if (!body.message_key) {
+      return 'message_key is required on /api/v1/chat/callback';
+    }
+    return null;
+  }
+  // 'legacy' transport keeps current implicit-dispatch behavior — no rules.
+  return null;
+}
```
IanLindsley: **[LOW]** Consider also rejecting `message_key` on the stream transport.

**Why:** `validateTransportFields` rejects `progress_callback_url`, `progress_mode`, and `progress_throttle_seconds` on `/chat/stream` — all the fields that would silently flip the legacy `/chat` endpoint into callback mode. But `message_key` is in the same logical family (it's the WhatsApp correlation ID and is required on `/chat/callback`), and it's currently allowed through on `/chat/stream` where it has no effect. That's a small foot-gun: a consumer that mistakenly thinks `message_key` does something on the streaming endpoint will get a clean 200 with no indication their field was ignored.

**Suggestion:** Either add
```ts
if (body.message_key !== undefined) {
  return 'message_key is not valid on /api/v1/chat/stream — use /api/v1/chat/callback';
}
```
or leave it as-is and document explicitly in the README that `message_key` is silently ignored on `/chat/stream`. The reject-and-explain path matches the existing behavior for the other three fields and seems more consistent — this PR is fundamentally about making the contract loud.

Not a blocker; happy to defer to your judgment if there's a reason `message_key` should be permissive.

IanLindsley: **[LOW]** `validateTransportFields` uses `!== undefined` for stream rejections but truthy checks for callback requirements — the asymmetry is worth a comment.

**Why:** On the stream branch, `body.progress_callback_url !== undefined` correctly rejects an explicit `null` (because `null !== undefined`) and an empty string. On the callback branch, `!body.progress_callback_url` accepts only truthy strings (rejects `null`, `''`, `undefined`). Both are correct for their use cases, but the asymmetry is non-obvious — a future contributor refactoring this might "fix" one to match the other and silently change validation semantics.

**Suggestion:** Add a one-line comment explaining the asymmetry, e.g.
```ts
// stream uses `!== undefined` so an explicit `null` is also rejected (we don't want clients to think null "unsets" a forbidden field);
// callback uses truthy so empty string and null also fail validation (we need a real URL).
```
No behavior change required.

tests/unit/chat-transport-dispatch.test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  it('rejects missing message_key', () => {
+    expect(
+      validateChatBody(
+        { ...baseBody, progress_callback_url: 'https://example.com/hook' },
+        'callback'
+      )
+    ).toBe('message_key is required on /api/v1/chat/callback');
+  });
+});
```
IanLindsley, line 137: **[LOW]** Missing test: `/chat/callback` with neither `progress_callback_url` nor `message_key`.

**Why:** The current callback tests cover "missing URL" and "missing message_key" individually but not "both missing". The behavior is deterministic (URL check fires first) but it's the implicit contract for one of the most common error paths — a consumer that doesn't read the docs at all will hit exactly this case. Documenting it as a test makes the precedence explicit and prevents an accidental future re-ordering.

**Suggestion:** Add
```ts
it('rejects when both progress_callback_url and message_key are missing', () => {
  expect(validateChatBody(baseBody, 'callback')).toBe(
    'progress_callback_url is required on /api/v1/chat/callback'
  );
});
```

Also, while you're here, consider adding one positive test for the legacy transport with `progress_callback_url` set (to lock in that legacy keeps allowing it). The current legacy-shared-rules test covers only the minimal body.

IanLindsley (commented):
## Code Review Summary

Solid, well-scoped additive change with clear motivation, a documented rollout plan, and good test scaffolding. The transport discriminator threads cleanly through `handleChatRequest` → `buildDOChatRequest` → DO `handleUnifiedChat`, the new `validateTransportFields` is short and readable, and logging context (`transport`, `delivery`) was added consistently throughout the new code paths. README updates are accurate against the implementation. The hono `4.12.7` → `4.12.12` bump and accompanying pnpm overrides are well-documented in the commit message; I verified no breaking API changes affect our usage patterns (`app.post`, `c.req.raw`, `c.req.json`, `c.json`, `app.use`) and that none of the patched code paths (`serveStatic`, `toSSG`, `ipRestriction`, cookie name routing) are used by this worker. `pnpm audit --prod` is clean after the bump.

I have **one Medium finding** that I would like addressed before merge, plus a few Lows that are nice-to-haves.

**Issues:** CRITICAL: 0 | HIGH: 0 | MEDIUM: 1 | LOW: 3

### Blocking Issues

1. **[MEDIUM]** `processQueueEntry` and `processSSEEntry`/`processCallbackEntry` still re-derive delivery from `body.progress_callback_url`, which silently desyncs from `transport` if the worker validation is ever bypassed. The PR description calls this out explicitly and says it's correct under current invariants — and it is — but the failure mode if anyone breaks the invariant is genuinely silent (callback-mode requests get a 202 then never deliver; SSE requests with no writer just drop their events). See the inline comment on `src/durable-objects/user-do.ts` line 240 for two concrete fix options. The cheap fix (DO-side defense-in-depth assertion mirroring the worker validation) is ~10 lines and unblocks merge; the robust fix (add `transport` to `InternalQueueEntry`) is a good follow-up.

### Definition of Done

This PR has no linked issue, so no formal DoD checklist. However, the PR description states a clear goal: "lock the transport explicitly". I checked that goal against the implementation:

- [x] `/api/v1/chat/stream` always SSE — verified in routing + `handleUnifiedChat` `isCallbackDelivery` derivation.
- [x] `/api/v1/chat/callback` always 202 + webhook — verified.
- [x] `/api/v1/chat/stream` rejects `progress_callback_url`/`progress_mode`/`progress_throttle_seconds` with 400 — covered by unit tests.
- [x] `/api/v1/chat/callback` requires `progress_callback_url` and `message_key` — covered by unit tests.
- [x] `/api/v1/chat` unchanged — verified by code inspection (no behavior change in the legacy branch).
- [x] Validation rejections logged at `warn` level with full context — `chat_validation_failed` includes `transport`, `error`, `user_id`, `client_id`. Good.
- [x] DO routing dispatches the new paths — covered by e2e smoke tests.
- [x] README accurately reflects code behavior — verified the table, the transport descriptions, and the rejected-fields note.
- [x] Version bumped to 2.13.0 — verified.
- [ ] DO defense-in-depth for transport invariant — see Medium above.

### Other Issues

All in inline comments:

- **[LOW]** `validateTransportFields` does not reject `message_key` on `/chat/stream`. The other three callback-flavored fields are rejected, so this is a small consistency gap — a consumer who sends `message_key` to `/chat/stream` thinking it does something will get a clean 200 with no signal. Either reject it or document the silent ignore in the README.
- **[LOW]** `validateTransportFields` uses `!== undefined` for the stream branch but `!body.field` (truthy) for the callback branch. Both are correct for their intent but the asymmetry is non-obvious — a one-line comment explaining why prevents an accidental future "fix" that changes semantics.
- **[LOW]** Unit test gap: no test for `/chat/callback` with both `progress_callback_url` and `message_key` missing. Behavior is deterministic but undocumented in tests. Also worth adding a positive legacy test that exercises `progress_callback_url` (current legacy test covers only the minimal SSE-style body).

### Positive Highlights

- **Clean discriminator threading.** `ChatTransport` is a small, well-named union and propagates through every layer (worker route → `handleChatRequest` → `buildDOChatRequest` → DO `handleUnifiedChat` → `enqueueAndReturn`) without polluting unrelated functions. The `doChatPathForTransport` helper centralizes the path mapping in one place.
- **Logging discipline.** Every new log site (`chat_validation_failed`, `chat_invalid_json`, `do_routed`, `chat_immediate`, `chat_enqueued`, `request_received`, `request_timing_summary`, `request_error`) was extended to carry the `transport` field. This is exactly the kind of observability discipline CLAUDE.md asks for, and it'll make production debugging across the three endpoints trivial.
- **Validation rejections are loud.** `chat_validation_failed` was added at the right severity (`warn`, since it's a client error) with the right context (`transport`, `error`, `user_id`, `client_id`). Easy to filter, easy to correlate.
- **The `delivery` vs `transport` distinction in DO logs** — `delivery: isCallbackDelivery ? 'callback' : 'sse'` plus the separate `transport` field — is a thoughtful ergonomic touch. It lets ops distinguish "consumer used the legacy endpoint" from "consumer used the explicit endpoint" while still seeing what mode the request was actually processed in.
- **Excellent rollout plan in the PR description.** The four-PR sequencing across three repos with explicit version targets is the right way to ship a contract change without breaking consumers, and the tracking issues for PR 2 and PR 3 are already filed. This is how contract migrations should be done.
- **Hono dep bump done right.** Direct dep AND pnpm overrides for the transitive (`@modelcontextprotocol/sdk` pulls hono and `@hono/node-server`) — that's the only way to get a clean audit. The commit message documents which advisories are addressed and explicitly notes that none of the vulnerable code paths are used by this worker. Excellent diligence.
- **E2E test framing.** The `posting bad JSON should hit handleUnifiedChat, not the Hono fallback 404` framing in the test file is a clever way to verify path dispatch without needing a fully wired chat pipeline. Smart use of the failure-case-as-positive-signal pattern.

### Recommendation

**NEEDS DISCUSSION → REQUEST CHANGES on Medium.**

The Medium finding is the only thing standing between this and a clean approval. I'm happy with the cheap fix (DO-side assertion), and once that's in I'll re-review and clear it. Lows can land in the same push or be deferred to a follow-up at your discretion.

Per CLAUDE.md, I will not approve via the review API (the agent runs as the PR author).

IanLindsley (commented):
## Follow-up Review (commit `ad3c984`)

This is a textbook fix-it pass. Every prior finding is resolved cleanly, the refactor improves the codebase rather than just patching the symptom, and no new issues of concern were introduced.

**Issues:** CRITICAL: 0 | HIGH: 0 | MEDIUM: 0 | LOW: 0 | OPTIONAL: 1

### Resolved

- **[MEDIUM]** Fragile transport invariant — **Fully resolved.** The new `parseChatBody` helper in `user-do.ts` runs `validateChatBody(body, transport)` as defense-in-depth at the top of `handleUnifiedChat`, before `enqueueAndReturn` is ever called. I traced the queue paths to confirm closure: `enqueueEntry` is the only writer to `QUEUE_KEY`, and it is called only from `enqueueAndReturn`, which is called only from `handleUnifiedChat`. With the new validation gate, no body can reach `processQueueEntry` / `processCallbackEntry` / `processSSEEntry` / `buildWebhookCallbacks` that doesn't satisfy the per-transport rules. The original failure modes (callback request returning 202 then never delivering, SSE request with no writer dropping events) are now provably impossible without also bypassing the DO. The defense-in-depth is genuine, not cosmetic.

- **[LOW 1]** `message_key` on `/chat/stream` — **Fixed.** `validateTransportFields` now rejects `message_key !== undefined` on the stream branch (lines 45–47 of `chat-validation.ts`), matching the treatment of the other three callback-flavored fields. New unit test `rejects message_key` confirms the rejection. Verified via the new legacy tests that `message_key` is still permitted on `/chat` (the `legacy callback mode` and `progress_mode + progress_throttle_seconds` tests both include `message_key: 'm1'` and assert `null`).

- **[LOW 2]** Asymmetric validation style comment — **Documented.** The new JSDoc on `validateTransportFields` (lines 24–33) explains the `!== undefined` vs truthy `!` asymmetry well, including the rationale ("any *presence* of a callback-flavored field on /chat/stream is an error" vs "empty strings are not usable as a URL or correlation key"). A future contributor refactoring this is now warned off the foot-gun.

- **[LOW 3]** Test gaps — **Closed.** Five new tests added (test count 12 → 17 in this file, suite total 388 → 393):
  - `/chat/stream rejects message_key`
  - `/chat/callback missing BOTH progress_callback_url and message_key (URL error first)` — locks in error precedence
  - `/chat legacy WITH progress_callback_url` — locks in legacy callback mode
  - `/chat legacy WITHOUT progress_callback_url` — locks in legacy SSE mode
  - `/chat legacy with progress_mode + progress_throttle_seconds` — locks in legacy permissive behavior

### New Issues

#### [OPTIONAL] `chat_validation_failed_in_do` log level

**Why:** The new DO-side validation logs at `warn` with event `chat_validation_failed_in_do`. That's the right level for a client error — but in production, this event firing means one of two things:

1. An internal caller bypassed worker validation (e.g., a test calling `stub.fetch('/chat/callback', ...)` directly with bad input) — expected, low signal.
2. Worker-side `validateChatBody` was broken by a future refactor and a real client request slipped through — this is a contract-level invariant violation that should be impossible.

The current log can't distinguish those two cases, and `warn` will get filtered alongside garden-variety client errors. The prior review's suggested event name `do_transport_invariant_violated` at `error` level was specifically about case 2. Worth considering whether to either (a) bump to `error` to make case 2 visible, or (b) split into two events, or (c) leave as-is and rely on the worker-side `chat_validation_failed` count being exactly equal to the DO-side `chat_validation_failed_in_do` count under the current invariant (a divergence detector at the dashboard level).

This is genuinely optional — the current code is correct and observable. Just noting that the failure-mode visibility is slightly dulled by the log level choice. Defer to your judgment.

### Positive Highlights

- **Right call to extract instead of duplicate.** Extracting `validateChatBody` into `src/utils/chat-validation.ts` and importing it from both `src/index.ts` (re-exported for tests) and `src/durable-objects/user-do.ts` is the right architectural move. Duplicating the validation logic to satisfy "DO defense-in-depth" would have created exactly the kind of drift risk the original Medium was about. One implementation, two callers, zero drift surface — this is how shared invariants should be enforced.

- **Clean module boundary.** `chat-validation.ts` imports only `type { ChatRequest, ChatTransport }` from `types/engine` — type-only, zero runtime coupling, zero circular-dep risk. Architecture cruise stays at 0 violations across 48 modules. The module's public API is minimal (only `validateChatBody`); `validateTransportFields`, `validateLanguageHint`, and `VALID_CHAT_TYPES` are correctly kept private. Onion architecture preserved.

- **`parseChatBody` discriminated-union return shape.** `{ body: ChatRequest; error?: never } | { body?: never; error: Response }` is exactly the right pattern. The `error?: never` on the success branch enforces mutual exclusion at compile time — a future refactor that accidentally returns both `body` AND `error` would fail typecheck rather than producing ambiguous runtime behavior. The narrowing at the call site (`if (parsed.error) return parsed.error; const { body } = parsed;`) is clean and safe. No casts, no `!`, no `as`.

- **Helper extraction kept `handleUnifiedChat` under the complexity budget.** Pulling parse + validate into a dedicated method instead of inlining is the right way to defend against the project's complexity-10 limit while still adding meaningful checks. The handler reads more clearly now too — the body parsing concern is fully isolated from the dispatch logic.

- **JSDoc on `parseChatBody`** explicitly explains *why* the re-validation exists ("re-checking here guarantees the transport → body invariant is enforced at the same place we rely on it for queue dispatch"). That's exactly the context a future maintainer needs to avoid "optimizing away" the apparent redundancy. Plus the updated comment in `handleUnifiedChat` now points back to the validation invariant. Good documentation discipline.

- **Test additions go beyond the minimum.** The "rejects missing both" test isn't just coverage — its name documents the error precedence as part of the contract. The three legacy tests collectively lock in that legacy is permissive and stays that way through any future refactor. This is exactly the kind of "tests as executable spec" the prior review was hoping for.

- **Commit message discipline.** The commit message walks through each finding by category (Medium, Low 1/2/3), explains the reasoning, and notes the test count delta. This is the kind of commit message that makes code archaeology a pleasure six months from now.

### Status

**4/4 issues resolved (1 Medium, 3 Low). 1 new Optional finding (log level — non-blocking).**

**Recommendation:** APPROVE. This is ready to merge — the Optional log-level finding can be addressed in a follow-up PR or deferred entirely. Per CLAUDE.md, I will not approve via the review API (the agent runs as the PR author).

IanLindsley:
Adding one more tracking issue found after merge — bt-servant-admin-portal also consumes `/api/v1/chat` as an SSE client via `worker/chat.ts:40`:

- bt-servant-admin-portal: https://github.com/unfoldingWord/bt-servant-admin-portal/issues/65 (migrate to `/chat/stream`)

Checked `baruch` too — not affected, it only calls bt-servant-worker's admin API (unchanged).

Issue #18, opened 2026-02-01
Improve QuickJS executor test coverage


IanLindsley:
## Description

From code review of PR #17, the following test coverage gaps were identified:

## Missing Test Cases

- [ ] Host function that throws during the call
- [ ] Multiple host functions called in parallel (only sequential chained awaits tested)
- [ ] Code that sets `__result__` multiple times (what's the expected behavior?)
- [ ] Very large results (what if `JSON.stringify` fails on the result?)
- [ ] Circular references in results
- [ ] Timeout during `processPendingCalls` (only tested during initial eval)

## Context

These are not blocking issues but would improve confidence in the executor implementation.

**Source:** Claude PR Review on #17

PR #120, opened 2026-04-14
feat!: flip /api/v1/chat to synchronous final-only JSON (v2.14.0)

IanLindsley:
## Summary

**BREAKING CHANGE.** `POST /api/v1/chat` now returns a synchronous final-only JSON response. The legacy body-dispatch behavior (SSE by default, webhook if `progress_callback_url` is present) is removed. The `/api/v1/chat/stream` and `/api/v1/chat/callback` sibling endpoints from v2.13.0 are unchanged and handle the SSE and webhook transports respectively.

This is the fourth and final PR in the four-PR rollout originally planned in PR #118 — the one Denis actually asked for.

## The new contract

| Endpoint | Transport | Response |
|---|---|---|
| `POST /api/v1/chat` | Synchronous JSON | `200 OK` with `{ message_id, responses, response_language, voice_audio_base64, voice_audio_url }`. `503 CONCURRENT_REQUEST_REJECTED` + `Retry-After: 5` when the user's DO is busy. Rejects `progress_callback_url`, `progress_mode`, `progress_throttle_seconds`, and `message_key` with 400. |
| `POST /api/v1/chat/stream` | SSE | Unchanged from v2.13.0 |
| `POST /api/v1/chat/callback` | Webhook | Unchanged from v2.13.0 |

## Why no queuing on `/api/v1/chat`

The synchronous transport holds the HTTP connection open for the duration of the orchestration. There is no way to enqueue a request and wait for an alarm to drain it while keeping the client's response stream alive — so when the user's DO already has the processing lock, we return `503` immediately and let the caller retry. SSE and callback transports queue cleanly (SSE via a live `TransformStream` writer, callback via an async closure that fires the webhook when ready); only `/api/v1/chat` has this constraint. Documented in README under "Concurrency note for `/api/v1/chat`".

## Code changes

- **`src/types/engine.ts`** — `ChatTransport`: `'legacy'` renamed to `'final'`.
- **`src/utils/chat-validation.ts`** — `validateTransportFields` refactored into two helpers, `forbidCallbackFields` (used by both `'final'` and `'stream'`) and `requireCallbackFields` (used by `'callback'`). Error messages now name the correct endpoint per transport.
- **`src/index.ts`** — `POST /api/v1/chat` route passes `transport='final'`. `doChatPathForTransport` maps `'final'` to `/chat/final` on the DO.
- **`src/durable-objects/user-do.ts`**
  - DO fetch dispatcher: `/chat` removed, `/chat/final` added. `/chat/stream` and `/chat/callback` unchanged.
  - `handleUnifiedChat`: the `isCallbackDelivery = transport === 'legacy' && !!body.progress_callback_url` branch is gone. Simple transport switch now.
  - New `processImmediateFinal`: awaits `processChat` synchronously and returns `Response.json({ message_id, ...chatResponse })`. Lock release + queue drain in `finally`.
  - Busy-path: `final` returns `503 CONCURRENT_REQUEST_REJECTED` with `Retry-After: 5`. Other transports still enqueue.

## Tests

- **`tests/unit/chat-transport-dispatch.test.ts`** — legacy test block deleted; replaced with a `'final'` transport block (minimal body accepted, all four callback-flavored fields each produce a 400). Group-chat `chat_id` requirement now exercised across all three transports. Suite: **395 → 398**.
- **`tests/e2e/chat-transport-dispatch.test.ts`** — legacy `/chat` DO-routing smoke test replaced with `/chat/final`. Stream and callback smoke tests unchanged.

## Local verification

- [x] `pnpm test` — 398/398 pass
- [x] `pnpm check` — clean
- [x] `pnpm lint` — 0 errors (5 pre-existing warnings unrelated to this PR)
- [x] `pnpm architecture` — 0 violations

## Rollout plan (DO NOT MERGE BLINDLY)

1. CI passes on this PR.
2. **Codex reviews this PR.** (Not the claude-code-review subagent — per the user's explicit instruction.)
3. Fix anything codex flags, re-push, re-review.
4. Merge → auto-deploy to **staging**. Verify end-to-end against staging consumers.
5. **GATE prod deploy** — do not trigger the manual Deploy workflow until all three consumer repos (`bt-servant-web-client`, `bt-servant-admin-portal`, `bt-servant-whatsapp-gateway`) have deployed their migrations to **prod**. The three consumer migrations are already on staging; prod status of each needs confirmation before this ships.
6. After prod deploys clean, verify `api.btservant.ai/health` reports `2.14.0` and run a curl smoke against `POST /api/v1/chat`.
7. Then — and only then — message Denis.

## Client impact summary

| Consumer | Needs | Status |
|---|---|---|
| bt-servant-web-client | `/chat/stream` | ✅ migrated to staging (issue #24) |
| bt-servant-admin-portal | `/chat/stream` | ✅ migrated to staging (issue #65) |
| bt-servant-whatsapp-gateway | `/chat/callback` | ✅ migrated to staging (issue #18) |
| bt-servant-telegram-gateway (Denis) | `/api/v1/chat` (synchronous JSON) | ⏳ blocked on this PR reaching prod |

## Related

- Parent rollout plan: PR #118 (v2.13.0, additive endpoints)
- Consumer tracking issues: #24, #65, #18 (all in sibling repos)
- Telegram thread with Denis dated 2026-04-13 where this contract was negotiated


IanLindsley:
## PR Review Status

### Critical
- None.

### High
- None.

### Medium
- None.

### Low
- None.

### Verdict
All issues addressed.

src/durable-objects/user-do.ts
```diff
@@ -346,6 +357,37 @@ export class UserDO {
...
+      const response = await this.processChat(body, workerOrigin, logger, timing);
+      logger.log('immediate_final_complete', { message_id: messageId });
+      return Response.json({ message_id: messageId, ...response });
+    } catch (error) {
+      const message = error instanceof Error ? error.message : 'Processing failed';
+      logger.error('immediate_final_error', error, { message_id: messageId });
+      return createErrorResponse('Processing failed', 'INTERNAL_ERROR', message, 500);
+    } finally {
+      await this.releaseLock();
+      await this.drainQueue(logger);
```
IanLindsley: `processImmediateFinal()` returns from the `try`, but this `finally` still awaits `drainQueue()`, so the HTTP response is not actually sent until every queued SSE/callback entry has finished processing. That breaks the new contract: one caller can end up waiting for unrelated queued work after its own orchestration is complete, and a backlog here can push the final request into worker timeouts. This path should release the lock and trigger queue draining without making the final response await the backlog.

src/durable-objects/user-do.ts
```diff
@@ -346,6 +357,37 @@ export class UserDO {
...
+  ): Promise<Response> {
+    const timing = createTimingContext();
+    try {
+      const response = await this.processChat(body, workerOrigin, logger, timing);
+      logger.log('immediate_final_complete', { message_id: messageId });
+      return Response.json({ message_id: messageId, ...response });
+    } catch (error) {
+      const message = error instanceof Error ? error.message : 'Processing failed';
+      logger.error('immediate_final_error', error, { message_id: messageId });
+      return createErrorResponse('Processing failed', 'INTERNAL_ERROR', message, 500);
```
IanLindsley: This catch collapses every downstream failure into `500 INTERNAL_ERROR`, but `processChat()` still throws `ValidationError` for user-input problems that are not covered by `validateChatBody()` (for example missing `message` for text requests or missing `audio_base64`/`audio_format` for audio requests; see `resolveMessageText()` below). On the new synchronous `/api/v1/chat` endpoint those malformed requests will now get a 500 instead of a 400, which makes client errors look retryable/server-side. Please preserve `ValidationError` as a 400 here or move the remaining request validation earlier.

IanLindsley (commented):
I found 1 high-severity issue and 1 medium-severity issue in the new `/api/v1/chat` final-response path.

IanLindsley:
Thanks for the review — both findings confirmed as real bugs. Fixed in `c99eae4`.

## HIGH — drainQueue blocking the caller

Moved `releaseLock` + `drainQueue` out of `processImmediateFinal`'s `finally` block and into a fire-and-forget background closure that launches after the Response is built but before it's returned. The caller now gets their JSON body the moment `processChat` resolves, regardless of what's queued behind it. Drain errors are caught and logged via the structured request logger with the message_id context.

This matches the pattern already used by `processImmediateSSE` and `processImmediateCallback` — those methods do the whole `processChat` + `release` + `drain` sequence inside a background closure and return the streaming/accepted response synchronously. For `final` I can't fire-and-forget `processChat` itself (the caller is waiting on its output), but I can fire-and-forget everything after it.

## MEDIUM — ValidationError → 500

Added an `instanceof ValidationError` branch to `processImmediateFinal`'s catch. `ValidationError` from `processChat` now returns `400 VALIDATION_ERROR` with the original message, logged at `warn` level (client-input problem, not a server fault). Mirrors the handling at `src/index.ts:986` (`handleChatRequest`).

`ValidationError` was already imported in `user-do.ts`; no new imports.

## Note on AudioTranscriptionError

Not handling it here — the worker-level `handleChatRequest` doesn't special-case it either (it falls through to 500). Keeping the DO consistent with the worker. If that class of failure needs to become a 400 across the board, happy to do that as a separate PR since it would touch both layers.

## Verification

- `pnpm check` — clean
- `pnpm lint` — 0 errors (5 pre-existing warnings)
- `pnpm test` — 398/398 pass
- `pnpm architecture` — 0 violations

Re-review when you have a minute.

IanLindsley (commented):
Re-review complete. The two previously reported issues in the final-response path are fixed in , and I did not find any new blocking issues in the follow-up patch.

IanLindsley:
Re-review note: the clean re-review applies to commit c99eae4. Both previously reported issues are fixed there, and I did not find any new blocking issues.

PR #121, opened 2026-04-14
fix(ci): close dev-dep audit bypass and clear vuln backlog (#119)

### LINKED ISSUE #119: CI security audit silently ignores dev-dependency vulnerabilities
## Summary

The CI `security-audit` job (`.github/workflows/ci.yml:28-32`) runs two audit steps, but only one of them actually gates the build. The second step, which audits **all** dependencies (including the dev tree), unconditionally exits 0 via `|| true`, so dev-dependency vulnerabilities — including High and Critical severity — are written to the CI log and then silently ignored. We also have no Dependabot / Renovate config, so dev deps drift stale over time with no automated bump pressure.

This was surfaced by an external security audit that flagged 32 dev-tree vulns against `origin/main` (12 unique packages, several High severity: `flatted` CVSS 8.9, `rollup` CVSS 8.8, `undici` with SSRF + header injection, etc.). CI was green the whole time because line 32 was eating the non-zero exit.

## Current state

```yaml
# .github/workflows/ci.yml

- name: Audit production dependencies
  run: pnpm audit --prod                              # HARD GATE — works correctly

- name: Audit all dependencies (informational)
  run: pnpm audit || true                             # ← THE HOLE
```

**What works:** the `pnpm audit --prod` hard gate caught the `hono 4.12.7` advisory on PR #118 and forced the bump to `4.12.12` before merge. The production-dep gate is doing its job.

**What doesn't work:** every other advisory against the dev tree is currently invisible to CI. The `|| true` was presumably added to avoid constant CI failures from transitive wrangler / vitest / eslint deps — but "noisy" was traded for "silent," and the `informational` output only gets read when someone emails a spreadsheet of findings.

**No automated bumps:** `.github/` contains only `workflows/` — no `dependabot.yml` or `renovate.json`. Dev deps drift unless someone manually runs `pnpm update`.

## Why this matters (even for a CF Workers project)

Dev deps do not ship to Cloudflare — the deployed artifact is a bundled JS file running on V8 isolates, so `undici`, `vite`, `rollup`, `esbuild`, etc. never exist at runtime. That's why the `|| true` was defensible on "these vulns don't reach prod."

But dev deps **do** run on the build host during `pnpm install`, `pnpm test`, and `pnpm run build`. A compromised transitive dep (see the recent `axios` supply-chain incident: malicious version executed a RAT at import time) running on a GitHub-hosted runner has access to:

- `CLOUDFLARE_API_TOKEN` / `CLOUDFLARE_ACCOUNT_ID` secrets used by `deploy.yml`
- `ENGINE_API_KEY` and `ANTHROPIC_API_KEY` exposed to the test job
- Arbitrary code execution on the build machine — which means it can arbitrarily modify the bundle before `wrangler deploy` ships it

So the real threat model for dev-dep vulns is **"supply-chain attack on the build host,"** not **"runtime exploit on the worker."** It's a different severity than production exposure, but it's not zero — and it's exactly the attack class the axios incident exercised.

### 1. Tighten the CI gate on line 32

Replace:

```yaml
- name: Audit all dependencies (informational)
  run: pnpm audit || true
```

with:

```yaml
- name: Audit all dependencies (High/Critical must be fixed)
  run: pnpm audit --audit-level=high
```

**Why `--audit-level=high` instead of removing `|| true`:** removing the flag entirely would fail CI on moderate/low dev-tree advisories, which are genuinely noisy for a Workers project. Setting the threshold at high/critical keeps the noise floor livable while still forcing a real response to anything serious. Production deps are already hard-gated regardless of severity by the step above.

This change alone will fail CI until the current backlog is cleaned up — which is intentional and correct.

### 2. Clear the existing backlog

Before (or in the same PR as) the gate change, bring the tree up to a clean state:

1. Run `pnpm update` to pick up minor/patch bumps automatically. Most of the flagged packages (`flatted`, `rollup`, `undici`, `minimatch`, `picomatch`, `brace-expansion`, `esbuild`, `defu`, `yaml`, `vite`) will resolve via transitive updates.
2. For anything still flagged where a patched version exists but a transitive parent pins it old, add a targeted override to `pnpm.overrides` in `package.json` — same pattern already used for `hono`, `@hono/node-server`, `express-rate-limit`, `path-to-regexp`.
3. Verify locally: `pnpm install && pnpm audit --audit-level=high` → should return clean.
4. Run the full verification suite: `pnpm test && pnpm lint && pnpm check && pnpm architecture`.

### 3. Add Dependabot for ongoing hygiene

New file `.github/dependabot.yml`:

```yaml
version: 2
updates:
  - package-ecosystem: "npm"
    directory: "/"
    schedule:
      interval: "weekly"
    open-pull-requests-limit: 5
    groups:
      dev-dependencies:
        dependency-type: "development"
        update-types:
          - "minor"
          - "patch"
      prod-dependencies:
        dependency-type: "production"
        update-types:
          - "minor"
          - "patch"
```

Effect: each week Dependabot opens **two** grouped PRs — one bundling dev-dep bumps, one bundling prod-dep bumps. Major version bumps stay manual (they land as individual PRs) because majors often require code changes. The grouped approach avoids PR spam while still keeping things fresh.

### 4. Document the intent

Add a short note to `CLAUDE.md` under `CRITICAL: Never Silently Swallow Errors` (or a new section) explaining that CI audits must fail-fast on high/critical in both trees, that Dependabot handles routine bumps, and that manual `pnpm update` is the fallback if Dependabot is down.

## Out of scope (for this issue)

- Switching off `pnpm audit` in favor of Snyk / Socket / GitHub Advanced Security. Maybe worth a separate spike if `pnpm audit` turns out to have coverage gaps in practice.
- Signing the deployed bundle or verifying wrangler's upload integrity. Separate concern.
- Pre-commit audit — ruled out upstream because it adds 5-10s to local commits. Stays as-is.

## Acceptance

- [ ] `.github/workflows/ci.yml` line 32 uses `pnpm audit --audit-level=high` (no `|| true`)
- [ ] `pnpm audit --audit-level=high` passes clean on the PR's HEAD
- [ ] `.github/dependabot.yml` committed with grouped dev/prod dep schedules
- [ ] Full verification suite (`pnpm test && pnpm lint && pnpm check && pnpm architecture`) clean
- [ ] CLAUDE.md (or README section) documents the audit gate policy
- [ ] Version bumped (patch)

## Priority

**High.** Not urgent enough to preempt the in-progress v2.14.0 transport rollout, but should land immediately after v2.14.0 ships. The gap has been open since the workflow was authored and no incident has exploited it, but "no incident yet" is not a security argument.

## Related

- PR #118 — where the prod-audit hard gate already demonstrated its value by catching the hono 4.12.7 bump.
- External audit (not tracked in GitHub) reporting 32 vulns on origin/main — stale findings, most already fixed or dev-only, but the underlying gate weakness is real.

IanLindsley:
## Summary

- Replaces `pnpm audit || true` with `pnpm audit --audit-level=high` so dev-tree high/critical advisories hard-fail CI instead of being silently logged.
- Clears the existing 32-vuln backlog: `pnpm update` for transitive patches, majors bump for `wrangler` (3→4), `@cloudflare/vitest-pool-workers` (0.12→0.14), and `vitest` (2→4), plus `undici >=7.24.0` / `vite >=6.4.2` in `pnpm.overrides`. Final `pnpm audit` reports **0 vulnerabilities** at any severity.
- Adds `.github/dependabot.yml` (weekly, grouped dev/prod minor+patch, limit 5) and documents the two-layer audit gate policy in `CLAUDE.md`.

Closes #119.

## Why

The `security-audit` job silently swallowed every high/critical dev-tree advisory via `|| true`, so advisories like `undici` 7.x (WebSocket DoS) and `flatted` (prototype pollution) were written to the log and ignored on every PR. Dev deps never reach the Workers runtime, but they **do** run on the CI build host with access to `CLOUDFLARE_API_TOKEN`, `ANTHROPIC_API_KEY`, and write access to the deploy artifact — the real threat model is supply-chain compromise of the build host (the axios incident attack class).

## Scope caveat

This PR ended up larger than the issue anticipated because the cleanest path to killing the remaining undici advisories was bumping the Cloudflare toolchain stack (`wrangler` / `vitest-pool-workers` / `vitest`) rather than force-overriding undici into miniflare 3/4 internals. The migration required:

- `vitest.config.ts`: rewrite from `defineWorkersConfig` (removed in pool-workers 0.14) to `defineConfig` from `vitest/config` + `cloudflareTest()` plugin
- `tests/e2e/env.d.ts`: update type reference to `@cloudflare/vitest-pool-workers/types`
- `tests/unit/orchestrator-separator.test.ts`: rewrite mock constructor from an arrow function to a named `function` (vitest 4 rejects arrow functions used as `[[Construct]]` targets)
- Add `vite ^6.4.2` as a direct devDependency so the `pnpm.overrides` entry actually applies (pnpm was otherwise picking vite 5 as a peer-dep auto-resolution)

## Verification (local)

- [x] `pnpm audit` → 0 vulnerabilities at any severity
- [x] `pnpm check` (tsc) → clean
- [x] `pnpm lint` → 5 pre-existing warnings, 0 errors
- [x] `pnpm architecture` → clean
- [x] `pnpm format:check` → clean
- [x] `pnpm test` → **418/418 passing across 22 files**
- [x] `wrangler build` under wrangler 4.72.0 → clean (ran via pre-commit hook)

## Test plan (staging, before merge)

Beyond CI — the toolchain bumps expand the risk surface beyond code-level checks, so please validate these manually against staging before approving the merge:

- [ ] CI `deploy.yml` workflow runs `wrangler deploy` cleanly against wrangler 4 (primary risk: CLI flag or config-parser changes)
- [ ] `pnpm dev` starts a local worker without errors under wrangler 4
- [ ] `UserDO` SQLite durable object behavior is unchanged under miniflare 4's rewritten storage layer — exercise read/write/session persistence via a real chat flow
- [ ] End-to-end chat flow against staging hits real Anthropic and does not regress (golden path + an error path to confirm error surfaces are still observable)
- [ ] No new warnings or deprecation notices in the staging deploy logs that weren't there on v2.14.1


ILindsley (approved):
Severity counts: critical 0, high 0, medium 0, low 0.

Verdict: approved after local diff review plus `pnpm run audit:prod`, `pnpm run audit:all`, `pnpm run check`, `pnpm run test`, and `pnpm run build` on the PR head.

ILindsley:
## PR Review Status

### Critical
- None.

### High
- None.

### Medium
- None.

### Low
- None.

### Verdict
All issues addressed.

PR #152, opened 2026-04-22
feat(modes): add published flag to prompt modes (#151)

### LINKED ISSUE #151: Add published flag to prompt modes with explicit backfill
## Summary

Introduce a `published: boolean` on `PromptMode` to distinguish draft modes from those visible to end users. Modes with `published !== true` are hidden from the orchestrator's `list_modes`/`switch_mode` tools; admin endpoints continue to return everything so the portal can manage drafts.

## Motivation

Today, every mode written to `PROMPT_OVERRIDES` KV is immediately visible to end users via `list_modes`. There is no way to work on a mode in the portal without it going live. We want explicit publish as a separate action from save.

## Data model

Extend `PromptMode` in `src/types/prompt-overrides.ts`:

```ts
export interface PromptMode {
  name: string;
  label?: string;
  description?: string;
  published?: boolean;   // treated as false when absent
  overrides: PromptOverrides;
}
```

Semantics: **`published === true` ⇒ visible to users. Anything else (false, undefined, missing) ⇒ draft.**

Because the read-side default is `false`, a one-time backfill is required before the filter ships — otherwise every existing mode disappears from end users.

## Rollout order

1. **Backfill (migration)** — script stamps `published: true` on every existing mode in `PROMPT_OVERRIDES` KV for `unfoldingWord` + `wordcollective` across staging + prod. Idempotent. Safe to run before any code changes because the current worker tolerates extra fields in stored JSON.
2. **Worker PR (this issue)** — type + validator + admin passthrough + filter + stale-mode handling. Merge → deploy to staging, verify → deploy to prod.
3. **Portal PR** — tracked in `bt-servant-admin-portal`. Depends on this.

## Scope

1. **Type + validator**: extend `PromptMode` and accept optional boolean `published` in `validatePromptMode`.
2. **Admin API passthrough**: `GET/PUT/DELETE /api/v1/admin/orgs/:org/modes[...]` (around `src/index.ts:473`) round-trip `published` unchanged. Admin endpoints return all modes — published and draft.
3. **Filter at the choke point**: in `src/durable-objects/user-do.ts:1340`, populate `availableModes` with `orgModes.modes.filter((m) => m.published === true)`.
4. **Stale selected-mode handling**: if a user's persisted `selectedMode` points at a mode that is no longer published (or no longer exists), fall through to org defaults gracefully rather than erroring. Verify `switch_mode` behavior in `src/services/claude/orchestrator.ts:874` against an unpublished name returns a clear \"unknown mode\" error to the model.
5. **Admin logging**: include `published` in `admin_action` payloads on PUT so publish/unpublish events are traceable.
6. **Backfill script**: add under `scripts/`. Reads `{org}:modes` from `PROMPT_OVERRIDES` KV for each org across staging + prod, stamps `published: true` on every existing mode, writes back. Idempotent.

## Known risk

Between the backfill landing and the worker PR landing, the **current** PUT handler may or may not preserve the new `published` field when round-tripping (it depends on whether it rebuilds the stored object from known keys only). If a mode is edited in that window, its `published` flag could be stripped. Mitigation: keep the gap short; staging validation will catch any regression before prod.

## Tests

- `validatePromptMode` accepts `published: true|false|undefined`, rejects non-boolean.
- UserDO populates `availableModes` with only `published === true` modes.
- `switch_mode` with an unpublished mode name errors from the orchestrator's perspective.
- A user whose `selectedMode` becomes unpublished resolves to org defaults without error on next turn.
- Admin `GET /modes` returns both drafts and published.

## Verification plan

- Local: unit tests above.
- Staging: after backfill, `GET /api/v1/admin/orgs/unfoldingWord/modes` — every mode has `published: true`. End-user chat sees same mode list as before.
- Staging: create a new mode via portal without publishing; confirm `list_modes` tool call does not include it. Publish it; confirm it appears.

IanLindsley:
Closes #151.

## Summary

Introduces `PromptMode.published?: boolean` so save and publish are separate actions. End users only see modes with `published === true`; the admin API still returns everything so the portal can manage drafts.

- **Type + validator**: `PromptMode.published?: boolean`; `validatePromptMode` accepts optional boolean, rejects non-boolean.
- **Round-trip fix**: `mergeExistingMode` now explicitly preserves `published`. Without this, a portal PUT that omitted the field would silently strip the flag from a previously-published mode.
- **Filter at the choke point**: `UserDO.buildModeContext` populates `availableModes` with `orgModes.modes.filter((m) => m.published === true)`. Both `list_modes` and `switch_mode` consume `availableModes`, so one filter covers both tools.
- **Stale selected-mode**: `resolvePrompts` treats a selected mode whose `published !== true` the same as "not found" — falls through to org defaults with a `mode_not_found` warn log carrying `reason: 'unpublished'`.
- **Admin log**: PUT `admin_action` now includes `published` for traceability.
- **Backfill script**: `scripts/backfill-mode-published.mjs` is idempotent, supports `--dry-run`, and handles staging+prod across both orgs.

Admin endpoints (`GET/PUT/DELETE /api/v1/admin/orgs/:org/modes[...]`) continue to return all modes regardless of `published`.

## Rollout (operator runbook)

The backfill must run **before** the worker deploy hits each environment, or every existing mode disappears from end users.

```sh
# 1. Staging dry-run (safe, read-only)
node scripts/backfill-mode-published.mjs --env=staging --dry-run

# 2. Staging real
node scripts/backfill-mode-published.mjs --env=staging

# 3. Merge → CI auto-deploys staging → verify (see below)

# 4. Prod dry-run
node scripts/backfill-mode-published.mjs --env=production --dry-run

# 5. Prod real
node scripts/backfill-mode-published.mjs --env=production

# 6. Prod auto-deploy proceeds → verify
```

Idempotent: any mode that already has `published` set (true OR false) is left untouched.

## Test plan

- [x] `pnpm test` — 440/440 passing (new validator + upsertMode tests)
- [x] `pnpm check`, `pnpm lint`, `pnpm build`, `pnpm architecture` clean
- [ ] Staging: backfill dry-run → real run → `GET /api/v1/admin/orgs/unfoldingWord/modes` shows `published: true` on every existing mode
- [ ] Staging: end-user chat — `list_modes` returns same set as before
- [ ] Staging: portal creates a new mode without publishing → `list_modes` does not include it; `switch_mode "<draft>"` returns "Mode '<draft>' not found"
- [ ] Staging: publish via portal → `list_modes` includes it; `switch_mode` succeeds
- [ ] Staging: unpublish a mode that a user has currently selected → next turn uses org defaults; logs show `mode_not_found` with `reason: 'unpublished'`
- [ ] Prod: repeat backfill + smoke checks after auto-deploy

## Notes

- Portal PR is tracked separately in `bt-servant-admin-portal`. Until it lands and starts sending `published: true`, any *new* mode created via the portal in the gap will be a draft (invisible) until re-saved with the updated portal.
- Patch version bump: backward-compatible optional field.


src/durable-objects/user-do.ts
```diff
@@ -1337,7 +1338,7 @@ export class UserDO {
     activeModeName: string | undefined
   ): ModeContext {
     return {
-      availableModes: orgModes.modes,
+      availableModes: orgModes.modes.filter((m) => m.published === true),
```
ILindsley, line 1344: `availableModes` is filtered here, but `activeModeName` is still passed through unchanged. If a user had already selected a mode that later becomes unpublished, `resolvePrompts()` now ignores that mode's overrides, but `handleListModes()` will still return it as `active_mode` because it only reads `modeContext.activeModeName`. That leaves the model in an impossible state where the active mode is not in the published list and can lead it to believe the draft mode is still selected. Please clear or mask `activeModeName` when the selected mode is missing or unpublished, and add a test around the stale-selection path.

ILindsley (changes requested):
1 medium issue found. The published-mode filter is correct, but stale selected modes are still surfaced as active in the tool context, so this PR does not fully fall back to defaults when a selected mode becomes unpublished.

ILindsley:
## PR Review Status

### Critical
- None.

### High
- None.

### Medium
- None.

### Low
- None.

### Verdict
All issues addressed.

ILindsley (approved):
Re-review: previous stale active_mode finding is fixed.  now masks unpublished or missing selections before they reach the tool layer, and the added tests cover the stale-selection cases. Findings: none.

ILindsley (approved):
Re-review: previous stale active_mode finding is fixed. resolveEffectiveMode() now masks unpublished or missing selections before they reach the tool layer, and the added tests cover the stale-selection cases. Findings: none.

Issue #68, opened 2026-02-18
Set up Dev/QA Env in CloudFlare for BT Servant V2


elsylambert:
Set up a dedicated Dev/QA environment in Cloudflare for BT Servant V2, separate from the current Production environment.

The existing Cloudflare configuration will remain as Production (Prod). This task focuses on creating a properly isolated Dev/QA environment to support testing, staging validation, and pre-release verification for V2.

PR #154, opened 2026-04-22
feat: admin chat requests bypass published-mode filter (#153)

### LINKED ISSUE #153: Admin chat requests should bypass the `published` filter on modes
## Context

#151 added a `published: boolean` on prompt modes and filtered `list_modes` / `switch_mode` / any end-user-facing mode lookup to `published === true`. That's correct for chat end users.

The portal (bt-servant-admin-portal #68 / PR #69) now exposes Draft vs Published so authors can iterate on a mode before exposing it. But the portal's built-in test chat pane uses the same `/api/v1/chat/stream` path that end users hit — so right now an admin cannot test a draft mode from inside the portal, which defeats the whole purpose of having drafts.

## What already exists

The admin portal's chat BFF (`worker/baruch.ts`) already attaches `is_admin: session.isAdmin` to the outbound chat request body. So bt-servant-worker already has a reliable signal that a given chat request originated from an admin-context session, not an end user.

## Ask

When an incoming chat request carries `is_admin: true`, do **not** apply the `published === true` filter to mode lookups on that request. Specifically:

- Resolving the user's currently-assigned mode should succeed even if it's a draft.
- `list_modes` invoked by the tool loop on an admin request should include drafts (or at least not hide the admin's currently-assigned one).
- `switch_mode` invoked on an admin request should accept draft mode names.

End-user (non-admin) requests must continue to see only published modes — no change there.

## Acceptance

- From the portal, an admin can pick a draft mode in the test chat pane, send a message, and the worker responds with the draft prompt stack applied.
- End users calling `/api/v1/chat/stream` without `is_admin: true` still cannot see or switch to drafts.
- Unit/integration test covering both branches.

## Related

- bt-servant-worker #151 (introduced `published`)
- bt-servant-admin-portal #68 / PR #69 (Draft/Published UI; blocked on this)

IanLindsley:
## Summary

- Adds optional `is_admin` on the chat request body. When `true`, the worker skips the `published === true` filter when resolving the user's current mode, when building `list_modes`' output, and when validating `switch_mode`.
- Non-admin requests are unchanged — end users still only see published modes.
- Unblocks the admin portal's built-in test chat pane (bt-servant-admin-portal #68 / PR #69), which reuses `/api/v1/chat/stream` and previously could not exercise draft modes.

### Changes

- `src/types/engine.ts` — add optional `is_admin` on `ChatRequest`.
- `src/utils/chat-validation.ts` — validate `is_admin` as optional boolean; extract `validateCoreFields` to keep complexity under the limit.
- `src/types/prompt-overrides.ts` — `resolveEffectiveMode` takes `options.includeUnpublished`.
- `src/durable-objects/user-do.ts` — thread `is_admin` through `resolvePrompts` (mode lookup + `mode_not_found` log) and `buildModeContext` (available modes surfaced to the orchestrator).
- Tests covering admin + end-user branches on `resolveEffectiveMode` and `is_admin` validation.

### Trust model

The `is_admin` flag is trusted at face value. The admin portal BFF (`bt-servant-admin-portal worker/baruch.ts`) is the only expected caller that sets it, and it's populated from the portal's server-side session. Defense-in-depth (shared secret / HMAC) is out of scope for this PR and can be layered later if needed.

## Related

- #151 introduced `published` on prompt modes
- bt-servant-admin-portal #68 / PR #69 (Draft/Published UI; blocked on this)

## Test plan

- [x] `pnpm lint`
- [x] `pnpm check`
- [x] `pnpm test` (452 passing)
- [x] `pnpm build`
- [ ] Staging: from portal test chat pane, pick a draft mode → send a message → verify draft prompt stack applied (`prompt_overrides_applied` log with `active_mode: <draft>`, `mode_overrides > 0`)
- [ ] Staging: end-user request (no `is_admin`) for the same org → `list_modes` does NOT include the draft; direct request of draft name masks to `reason: 'unpublished'`

Closes #153


IanLindsley:
## PR Review Status

### Critical
- None.

### High
- None.

### Medium
- None.

### Low
- None.

### Verdict
All issues addressed.

src/types/engine.ts
```diff
@@ -59,6 +59,16 @@ export interface ChatRequest {
...
 
+  /**
+   * Set to true by the admin portal BFF (bt-servant-admin-portal worker/baruch.ts)
+   * when the chat originates from an authenticated admin session. Bypasses the
+   * published-mode filter so authors can test draft modes from the portal's
+   * built-in test chat pane. Trusted at face value: the portal BFF is the only
+   * expected caller to set this, and it's populated from the portal's
+   * server-side session.
+   */
+  is_admin?: boolean;
```
IanLindsley: `is_admin` is being added as a regular request-body field and the worker only type-checks it before forwarding the body into mode resolution. On the `/api/v1/chat/stream` path, that means any caller can send `{ "is_admin": true }` and bypass the published-mode filter for `resolveEffectiveMode`, `list_modes`, and `switch_mode`. This is effectively an authorization bypass for draft modes. Please derive the admin bit from a worker-verified credential/header instead of trusting client JSON here.

IanLindsley (commented):
Blocking issue identified: the new `is_admin` path trusts a client-controlled request field as authorization. Any caller that can hit `/api/v1/chat/stream` can opt into draft-mode visibility and switching by sending `is_admin: true`. This needs a worker-verified signal before the published filter is bypassed.

IanLindsley:
Addressed in 7dbe3ff.

The admin signal is no longer a body flag. Instead, the worker derives it from `client_id` via a new `isAdminClient()` helper, with `client_id === "admin-portal"` as the only admin-capable value today. The portal BFF already hardcodes `CLIENT_ID = "admin-portal"` in its own source (bt-servant-admin-portal `worker/baruch.ts:5`) and ships it on every chat relay, so no portal change is needed — the `is_admin` field we were adding was redundant with a signal already on the wire.

Changes:
- `is_admin` removed from `ChatRequest`.
- `validateChatBody` now rejects any `is_admin` in the body with a clear error, so older clients that try to self-assert see a 400 rather than a silent no-op.
- `ADMIN_CLIENT_IDS`/`isAdminClient()` live in `src/utils/chat-validation.ts` with a comment on the trust model.
- `UserDO` derives the admin bit from `client_id` in both `resolvePrompts` and the `buildModeContext` call site.
- Tests cover the rejection path and the `isAdminClient` allowlist.

Residual risk: any caller that already holds `ENGINE_API_KEY` can also set `client_id: "admin-portal"`. That caller is already an authenticated internal service with full API access, and tighter per-client auth is out of scope for this PR — but the specific bypass of "anyone flipping a JSON bit" is closed.

Ready for re-review.

PR #165, opened 2026-04-23
feat(prompt): enforce canonical media markdown and ban URL hallucination (#163)

### LINKED ISSUE #163: Constrain Claude to canonical media markdown — stop inventing URLs, force ![](img) / [](video) shapes
## Summary

The worker emits media URLs in inconsistent markdown shapes, and occasionally invents URLs the tool never returned. Both downstream clients (`bt-servant-web-client`, `bt-servant-whatsapp-gateway`) depend on a tight contract to render media inline. We need the LLM to stick to canonical shapes, and we need the rules to be **non-overridable** — they're a rendering protocol, not a style preference.

## Background / Related Work

- **Web client `v1.7.1`** (`unfoldingWord/bt-servant-web-client#32`, tag `v1.7.1`) — just shipped. Renders:
  - `alt` → inline `` (react-markdown default)
  - `[label](url.mp4)` → inline `<video>` via custom remark plugin (first-occurrence dedup)
- **WhatsApp gateway** — in-flight work adds a markdown-unwrap pass (`unwrapMediaMarkdown`) so after URL extraction the caption reads naturally (`Mount Tabor Map` instead of `Mount Tabor Map`).

Both clients assume the worker will emit one of these two shapes per media URL. Today it doesn't.

### 1. Hallucinated video URL (Mount Tabor)

- Request: `aa1e17fa-8f8e-4b0a-9ebb-9298c1628ebf` on `bt-servant-worker-staging`, 2026-04-23T19:07:37Z
- Prompt path: FIA video lookup for Matt 17:1-13 (Transfiguration)
- Tool result from `fia_get_pericope_media`: `Video:` field was **empty** for Mount Tabor
- LLM nonetheless emitted:
  ```
  **▶️ [Watch the Mount Tabor Video](https://s3.amazonaws.com/cbbt-er.public/media/videos/a109/720p.mp4)**
  ```
- URL was pattern-constructed from the sibling `photos/a109/…` path; `videos/a109/720p.mp4` returns 404. Web client now faithfully renders a `<video>` pointing at a 404 → broken player placeholder.

### 2. Inconsistent image formatting

Same response, accompanying media:
```
- 🗺️ [Mount Tabor Map](https://…/maps/…/c104/v1/1500.jpg)
- 📸 [Photo of Mount Tabor](https://…/photos/a109/1500.jpg)
- 📸 [People Kneeling (reverence/worship)](https://…/photos/a667/1500.jpg)
```
Emitted as `[text](url.jpg)` (link syntax) instead of `alt` (image syntax). Web client renders them as clickable text links rather than inline images. Earlier responses for other prompts (e.g., "show me a picture of pottery") correctly used `alt`. The worker is inconsistent.

### Non-overridable constant

Rendering format is a protocol contract with downstream clients, not a per-org style preference. Adding a new overridable slot would leave it clobberable by well-meaning org admins — same failure mode we're trying to fix. Put the rules as a non-overridable constant concatenated into the system prompt in `src/services/claude/system-prompt.ts`, with a comment explaining why.

Sketch:

```ts
// src/services/claude/system-prompt.ts

// Non-overridable rendering contract. Downstream clients (web, WhatsApp)
// parse Claude's output for media URLs and render them natively — this rule
// set is the wire format between the worker and those clients. Do not move
// this into a prompt-override slot: an org accidentally clobbering it breaks
// inline media rendering across every conversation for that org.
const MEDIA_FORMATTING_RULES = \`## Media URL formatting (REQUIRED)

When referencing a URL returned by a tool, use EXACTLY these formats:

- **Image URLs** ending in \`.jpg\`, \`.jpeg\`, \`.png\`, \`.webp\`, or \`.gif\` → use markdown image syntax: \`![descriptive alt text](url)\`. NEVER use link syntax \`[text](url)\` for an image URL. NEVER emit a bare image URL.
- **Video URLs** ending in \`.mp4\`, \`.webm\`, \`.mov\`, \`.m4v\`, or \`.ogv\` → use markdown link syntax: \`[descriptive label](url)\`. NEVER wrap a video URL with \`!\` (never \`![…](url.mp4)\`). NEVER emit a bare video URL.

## Never invent URLs

You may ONLY reference media URLs that were explicitly present in a tool's return value.
- If a tool result has an empty or missing field for a media item (e.g., \`Video:\` with no URL), DO NOT emit a reference for that item. Skip it or say \"no video available for this passage.\"
- NEVER construct a URL by pattern-matching from another field (e.g., do not derive a video URL from a photo URL by swapping path segments like \`photos/…\` → \`videos/…\`).
- Copy URLs verbatim from tool output.\`;
```

Inject it in the existing assembly order (currently in `system-prompt.ts` around line 136–141) at a position where it follows the slot-based content — so slot-level overrides run first, and this contract can't be overridden by them.

### Why option 4 (non-overridable) over option 3 (new slot)

| | Overridable slot | Non-overridable constant |
|---|---|---|
| Contract durability | An admin can `PUT` empty string → clients break | Can't be broken via admin API |
| Blast radius of accidental clobber | Every conversation for that org | None |
| Semantic fit | \"Output format\" isn't a style preference | Matches — this is a protocol layer |
| Escape hatch if future client needs different format | Admin-UI edit (wrong level — needs code review) | Branch on client capability in code (right level) |
| Code surface added | PROMPT_OVERRIDE_SLOTS entry + interface field + DEFAULT_PROMPT_VALUES entry + validation + admin UI row | Single `const` + injection line |

## Acceptance Criteria

- [ ] Every response that references an image URL uses `alt` syntax.
- [ ] Every response that references a video URL uses `[label](url)` syntax (never ``).
- [ ] Response does NOT reference a media item whose tool-returned URL is empty or missing.
- [ ] Response does NOT contain a media URL that was not present verbatim in a tool's return value.
- [ ] Rules apply regardless of which org / mode / user prompt-override stack is in effect (non-overridable).

### Unit / integration
- Feed a canned `fia_get_pericope_media` result where `Video:` is empty; assert the orchestrator output contains no video link for that item.
- Feed a canned result with valid image + video URLs; assert output uses `` for images and `[](...)` for videos (no cross-syntax mixing).
- Construct an adversarial result missing a video but with a sibling photo URL; assert the LLM does not invent a video URL from the photo path.

### Live QA (staging)
- Repeat the Mount Tabor Transfiguration prompt that triggered this incident; confirm no broken video placeholder.
- FIA Matthew video prompt (all pericopes); confirm every video is emitted as `[label](url.mp4)`.
- Aquifer VideoBibleDictionary prompt (e.g., Garden Tomb, Gethsemane); confirm the existing linked-thumbnail pattern still works end-to-end.

## Related

- `unfoldingWord/bt-servant-web-client#32` — inline `<video>` rendering + first-occurrence dedup plugin (v1.7.1)
- WhatsApp gateway `unwrapMediaMarkdown` work (separate PR, in progress)
- #158 — silent end-turn (adjacent methodology rule; both are response-quality guardrails)

## Impact

**Severity**: Medium. Today produces visibly broken media embeds (Mount Tabor 404, inline images shown as links) on both clients whenever the LLM picks the wrong shape. Frequency scales with how often we emit media-heavy FIA / Aquifer responses — rising.

IanLindsley:
## Summary

- Downstream clients (web v1.7.1, WhatsApp gateway) render media inline only when Claude emits media URLs in canonical shapes: `alt` for images, `[label](url)` for videos. The worker was inconsistent (sometimes emitting `[text](url.jpg)` link syntax for images) and occasionally invented video URLs by pattern-matching sibling photo paths — staging hit a Mount Tabor 404 where `fia_get_pericope_media` returned an empty `Video:` and the LLM fabricated a `videos/a109/720p.mp4` URL.
- Adds `MEDIA_FORMATTING_RULES` as a non-overridable constant in `system-prompt.ts`, appended just before the closing slot so slot-level overrides can't displace it. Deliberately NOT a `PromptSlot` — an org clobbering a rendering contract would silently break every media response for that org.
- Covers image syntax, video syntax, and a never-invent-URLs guardrail (copy verbatim from tool output, skip when tool field is empty, no path-swapping).

## Test plan

- [x] `pnpm test tests/unit/system-prompt.test.ts` — 4 new tests assert rules presence, non-overridability under full slot overrides, and placement after memory + audio guidance, before closing
- [x] `pnpm test` — 460 tests pass
- [x] `pnpm check` — tsc clean
- [x] `pnpm lint` — no new warnings
- [ ] Staging QA after deploy: repeat Mount Tabor / Matt 17:1-13 Transfiguration prompt — no invented video URL, no broken `<video>`; images render inline on web client (not as text links)
- [ ] Staging QA: generic FIA pericope with valid image+video — images ``, video `[]()`
- [ ] Staging QA: Aquifer VideoBibleDictionary prompt (Garden Tomb / Gethsemane) — existing linked-thumbnail pattern unregressed

Closes #163


src/services/claude/system-prompt.ts
```diff
@@ -166,6 +186,7 @@ export function buildSystemPrompt(
     sections.push(VOICE_RESPONSE_GUIDANCE);
   }
   sections.push(...buildConditionalSections(preferences, history));
+  sections.push(MEDIA_FORMATTING_RULES);
```
IanLindsley: `MEDIA_FORMATTING_RULES` is now unconditional, but voice requests set `isVoiceMessage=true` and `VOICE_WRITING_RULES` explicitly says "Do NOT use any markdown formatting". That conflict is reachable in production (`user-do.ts` sets `isVoiceMessage` for every `message_type === 'audio'`), and the TTS path then strips markdown links/images entirely before synthesis. So a voice response that includes media will either violate voice mode or degrade into bare alt text/labels with the URL lost. This should be gated off for voice mode, or the voice-mode guidance needs a different media strategy plus a regression test covering `isVoiceMessage: true`.

IanLindsley (commented):
1 medium issue found: the new media-formatting contract is applied even for voice messages, where the existing prompt explicitly forbids markdown and the TTS path strips markdown out of the final response.

IanLindsley:
## PR Review Status

### Critical
- None.

### High
- None.

### Medium
- None.

### Low
- None.

### Verdict
All issues addressed.

IanLindsley (commented):
All issues addressed.

Issue #158, opened 2026-04-23
Claude ends turn silently after 'let me check X' narration — user sees 3-min gap


IanLindsley:
### Summary

When Claude narrates a next step ("let me try the entity profile directly and also check the FIA pericope media") and then executes tool calls in subsequent orchestration iterations, the turn frequently ends with `stop_reason: end_turn` and **no text content block**. The user never sees a closing summary — just the "let me check X" narration, then silence. From the user's perspective the system looks broken or hung.

This is a prompt-methodology issue, not an infrastructure issue. Fixing it likely belongs in the `methodology` and/or `instructions` prompt-override slots for the default `unfoldingWord` org and the published coaching modes.

### Observed Incident

- Request: `b786626e-17f2-404c-965c-f94b880250c6` on `bt-servant-worker-staging`, 2026-04-23
- User (WhatsApp): *"hmmmm. but it does. it serves up links."* (at 13:54:50 UTC)
- BT Servant replied with three short status messages. Last webhook_send at **13:55:41 UTC** with text:
  > "That article is rich text with links but no images. Let me try the entity profile directly and also check the FIA pericope media which is the actual source of visual content."
- Orchestration continued through iterations 5–9 (13:55:41 → 13:55:56):
  - `aquifer.entity person:James` → AbortError (30s upstream timeout)
  - `fia_list_languages([])` → invalid args (see #156)
  - Retry succeeded; subsequent `fia_list_books` + `fia_get_pericopes` fetched data
- **13:55:56 UTC** — `claude_response stop_reason: end_turn`, **no text block**, `process_chat_complete total_ms: 65861, response_count: 3`. The iteration 9 text response was empty.
- User waited ~3 minutes in silence, then at **13:58:59 UTC** typed: *"did you die? what happened? still waiting!!"*

### Why This Happens

Current methodology-style prompt encourages Claude to speak progress messages like "let me try X". When X then completes (or partially completes) in subsequent tool-use iterations, Claude evidently considers the turn complete without summarizing findings back to the user. The `stop_reason` transitions from `tool_use` to `end_turn` after the last tool result is consumed, with no requirement to emit a concluding text block.

This is especially likely when:
- The intermediate tool calls hit errors/retries that Claude self-recovers from (so the "happy path" to a summary is long)
- The narrated action was exploratory ("let me check X") without a concrete expected output

### Proposed Fix (Prompt-Override Level)

[...]

PR #167, opened 2026-04-23
chore(prompt-overrides): raise MAX_OVERRIDE_LENGTH to 8000

IanLindsley:
## Summary

- Bumps the per-slot `MAX_OVERRIDE_LENGTH` validator from 4000 → 8000 in `src/types/prompt-overrides.ts`.
- The 4000-char limit was rejecting appends to existing org/mode `closing` slots that are already at/near the limit (e.g. `translation-coach` is 4037, `fia-coach` is 3950 — both legacy data set before this validator was tightened, or via paths that bypassed it).
- Discovered while applying the prompt-override fix from #158 ("Closing the Turn" rule) — appending the new guidance to existing `closing` slots was blocked.

## Coordination

A companion bump is needed in `bt-servant-admin-portal` (`MAX_SLOT_LENGTH` in `src/types/prompt-override.ts`) so the editor UI doesn't block at 4000 chars while the API would accept up to 8000. Filing a separate issue there.

## Test plan

- [x] `pnpm check` — passes
- [x] `pnpm lint` — passes (warnings unchanged)
- [x] `pnpm test` — 461 / 461 pass; `tests/unit/prompt-overrides.test.ts` references the constant by name, so the boundary tests still validate against the new limit
- [x] `pnpm build` — passes
- [ ] Post-deploy: PUT existing org `closing` + the three coaching modes' `closing` slots with the #158 "Closing the Turn" appended text on staging → confirm 200, then prod


IanLindsley:
## PR Review Status

### Critical
- None.

### High
- None.

### Medium
- None.

### Low
- None.

### Verdict
All issues addressed.

IanLindsley (commented):
No findings. I traced the `MAX_OVERRIDE_LENGTH` change through validation, mode handling, KV persistence, and system-prompt assembly, and verified the targeted prompt-override tests still pass locally.

Issue #173, opened 2026-04-29
Explicit fonts[] in ptxprint payloads (vs bundled Charis SIL)


ILindsley:
## Background

In the v1 of the ptxprint-mcp integration (branch `feat/ptxprint-mcp-integration`), payloads are submitted with `fonts: []`. The ptxprint-mcp container falls back to bundled Charis SIL — the minimal smoke (`smoke/minimal-payload.json` in ptxprint-mcp) confirms this works.

Per discussion with a ptxprint creator, sending explicit font references and language codes may be the preferred pattern. Language codes are already covered (per-request `Settings.xml` interpolation). Fonts are not.

## Scope

1. Confirm with the ptxprint creator whether explicit fonts are required or recommended.
2. One-time setup: download Charis SIL TTFs (Regular/Bold/Italic/BoldItalic) from SIL's release page, upload to our `PTXPRINT_BUCKET` at a stable prefix like `fonts/charis-sil-<version>/`, record sha256s.
3. Wire each preset's `fonts[]` to reference those four entries (`{ family_id, version, filename, url, sha256 }` per the FontSchema in ptxprint-mcp's `src/payload.ts`).
4. Each preset's `ptxprint.cfg` already specifies `fontregular = Charis SIL|...` so PTXprint will pick up the supplied TTFs automatically.

## Reference

`smoke/fonts-payload.json` in ptxprint-mcp shows the exact pattern (with Gentium Plus 6.200) — uploaded TTFs in R2, referenced by URL + sha256 in the payload.

## Files / entry points

- `src/services/ptxprint/presets/` — each preset gets a `fonts[]` array added
- `src/services/ptxprint/payload-builder.ts` — wire fonts through to the payload

Issue #172, opened 2026-04-29
Integrate ptxprint-mcp config-generator tool (replace hardcoded presets)


ILindsley:
## Background

In the v1 of the ptxprint-mcp integration (branch `feat/ptxprint-mcp-integration`), bt-servant-worker ships with three hardcoded layout presets stored as TS string constants under `src/services/ptxprint/presets/`:

- `paperback-a5`
- `letter-2col`
- `large-print-a4`

A colleague is adding a fourth tool to ptxprint-mcp that returns information for generating arbitrary configs on the fly.

## Scope

Once that tool ships:

1. Wire bt-servant-worker to call the new tool from inside `generate_scripture_pdf` (and possibly expose it as another internal helper or just let Claude reach it via the catalog).
2. Replace the three hardcoded presets with on-the-fly config generation driven by user intent (page size, columns, font size, etc.).
3. Decide whether to keep the preset names as a thin compatibility wrapper ("paperback-a5" → call config-generator with the equivalent params) or remove them entirely.

## Files / entry points

- `src/services/ptxprint/presets/` — current home of the three presets
- `src/services/ptxprint/macro-tool.ts` — `generate_scripture_pdf` entry point
- `src/services/ptxprint/payload-builder.ts` — where `config_files` gets assembled into the payload

## Blockers

- New ptxprint-mcp tool name + schema (TBD)

PR #174, opened 2026-04-29
feat(ptxprint): integrate ptxprint-mcp for scripture PDF generation

IanLindsley:
## Summary

Adds end-to-end pipeline for generating print-ready scripture PDFs via the `ptxprint-mcp` Cloudflare Worker. Exposes both a macro tool (`generate_scripture_pdf`) for the happy path and a helper (`prepare_usfm_source`) so Claude can drop down to raw `submit_typeset` / `get_job_status` / `cancel_job` from the catalog when needed.

- New macro and helper internal tools wired through the existing `request_audio` dispatch pattern in the orchestrator.
- Three hardcoded layout presets (`paperback-a5`, `letter-2col`, `large-print-a4`) with cfg/sty/`Settings.xml` templates; default is `paperback-a5`.
- USFM resolved from Door43, hashed, re-hosted in our R2 (`PTXPRINT_BUCKET`) under content-addressed keys; PDFs mirrored from ptxprint's R2 to ours so user-facing URLs stay on our domain.
- New unauthenticated `/public/ptxprint/*` route streams R2 objects so the ptxprint container (and downstream WhatsApp/web consumers) can fetch without auth, registered before the `/api/*` auth middleware.
- Bounded sync polling: 5s interval, 60s ceiling. Timeouts return `pending` + `job_id` so the user can re-ask (cache hits) or cancel via the raw `cancel_job` tool.
- `ChatResponse` extended with `attachments[]`; `AttachmentsContext` side-channel mirrors the existing `AudioContext` pattern in `UserDO`.
- Comprehensive structured logging on every external boundary (DCS fetch, R2 ops, MCP calls, polling iterations) with start/complete pairs and rich context. No silent catches.
- 6 new unit test files cover presets, payload builder, polling, USFM resolver, PDF mirror, and macro tool end-to-end with all I/O mocked.

## Out of scope (tracked as follow-ups)

- Config-generator MCP tool integration (#172)
- Explicit `fonts[]` in payloads (#173) — v1 uses bundled Charis SIL
- WhatsApp gateway document delivery — issue filed in `bt-servant-whatsapp-gateway`
- Web client PDF rendering — issue filed in `bt-servant-web-client`

Per-org registration of `ptxprint-mcp` in `MCP_SERVERS` KV is a follow-up admin step, consistent with how `translation-helps-mcp` is provisioned today.

## Test plan

- [x] `pnpm check` (TypeScript) clean
- [x] `pnpm architecture` (no circular deps) clean
- [x] `pnpm lint` clean (warnings only, no new ones)
- [x] `pnpm test` — 512 tests passing locally including 6 new ptxprint test files
- [ ] Register `ptxprint-mcp` in default org `MCP_SERVERS` via the admin endpoint
- [ ] Curl-based staging smoke per the plan:
  - 3 presets + 1 default — confirm `attachments[0]` points at our `/public/ptxprint/pdfs/...` URL, not `klappy.workers.dev`
  - Fetch the PDF URL with NO `Authorization` header — must return 200 (Meta WhatsApp will fetch unauthenticated)
  - Fetch the corresponding USFM URL unauthenticated — must return 200 (proves the container could reach it)
  - Inspect Cloudflare Worker logs via the `cloudflare-logs` skill — confirm `generate_scripture_pdf` dispatched, DCS + `submit_typeset` + `get_job_status` traces present, no swallowed errors
  - Repeat one call to verify ptxprint-side cache hit (`cached: true`)
- [ ] Cancel-path smoke — get back a `pending` state with `job_id`, follow up with "cancel that" — confirm Claude calls raw `cancel_job` from the catalog
- [ ] Raw-tool smoke — explicitly ask Claude to use `submit_typeset` directly — confirm it calls `prepare_usfm_source` first, then `submit_typeset`, then polls `get_job_status` to completion


src/durable-objects/user-do.ts
```diff
@@ -1332,6 +1327,46 @@ export class UserDO {
...
+    const voiceAudioUrl = audioKey ? audioKeyToUrl(audioKey, workerOrigin) : null;
+    const attachments = attachmentsContext.list();
+    // prettier-ignore
+    logger.log('process_chat_complete', { total_ms: Date.now() - startTime, response_count: responses.length, has_voice_audio: voiceAudioUrl !== null, voice_audio_key: audioKey, total_response_chars: responses.join('').length, attachment_count: attachments.length, attachment_summary: attachments.map((a) => ({ type: a.type, filename: a.filename, size_bytes: a.size_bytes })), response: responses.join('\n') });
+    return {
+      responses,
+      response_language: effectivePreferences.response_language,
+      voice_audio_base64: null,
+      voice_audio_url: voiceAudioUrl,
+      ...(attachments.length > 0 ? { attachments } : {}),
```
IanLindsley, line 1383: Medium: attachments are only surfaced on the immediate `ChatResponse`; they are not persisted into chat history. `saveConversation()` still stores only `assistant_response`, `voice_audio_key`, and `speaker`, and the history response types do not expose attachments.

Why it matters:
After a reload or `/history` fetch, previously generated PDFs disappear from the conversation even though they were part of the assistant response.

Suggested direction:
Persist attachment metadata with each history entry and expose it from the history API the same way `voice_audio_url` is reconstructed today.

src/types/engine.ts
```diff
@@ -100,6 +116,13 @@ export interface ChatResponse {
 
   /** URL to fetch the audio from R2, or null if no audio was generated */
   voice_audio_url?: string | null;
+
+  /**
+   * Tool-produced artifacts (e.g., generated PDFs). Empty / omitted when no
+   * artifact-producing tool ran. Consumers that don't understand the type
+   * should ignore the field — never present a raw URL string in chat.
+   */
+  attachments?: Attachment[];
```
IanLindsley, line 125: High: `ChatResponse` now exposes `attachments[]`, but the callback transport still only serializes text and audio in `createWebhookCallbacks` / `ProgressCallbackSender.sendComplete`. Any client using `/api/v1/chat/callback` will never receive the generated PDF URL, so the new ptxprint flow is unusable on that transport.

Why it matters:
The feature appears to work on synchronous responses, but callback-based consumers lose the primary artifact entirely.

Suggested direction:
Extend the callback payload contract, sender, and tests to forward `attachments[]` on the final event.

IanLindsley:
Reviewed PR #174 focusing on correctness, transport integration, and client-visible artifact behavior.

Findings:
- High: 1
- Medium: 1

Top issues:
1. High: callback/webhook transport still drops `attachments[]`, so callback clients cannot receive the generated PDF URL.
2. Medium: generated attachments are not persisted into chat history, so a history fetch or client reload loses the PDF artifact for prior messages.

Inline comments were added on the relevant files. Residual limitation: this review was based on the diff and surrounding code paths; I did not run the PR branch.

IanLindsley:
Re-reviewed the latest PR #174 diff, focusing on the two previously reported issues around callback attachment delivery and history persistence.

Result:
- Critical: 0
- High: 0
- Medium: 0
- Low: 0

The callback transport now forwards `attachments[]` on the final webhook event, and attachment metadata is now persisted into chat history and returned on history reads. I did not find any new regressions in the updated diff.

Good for merge.

src/services/mcp/streamable-http-client.ts
```diff
@@ -0,0 +1,244 @@
...
+  logger.log('mcp_tool_call_start', {
+    server_id: server.id,
+    transport: 'streamable-http',
+    tool_name: toolName,
+    args: summarizeArgs(args),
+  });
+  let opened: ConnectedClient | null = null;
+  try {
+    opened = await openClient(server);
+    const raw = (await opened.client.callTool({
```
ILindsley, line 244: Medium: the new `streamable-http` tool path appears to bypass the existing max-response-size guardrails. `CallMCPToolOptions.maxResponseSizeBytes` is honored in the JSON-RPC path via `buildToolCallSendOptions`, but this SDK call ignores `options?.maxResponseSizeBytes` entirely and returns the tool result unbounded. That is a transport-parity regression: a streamable-http server can now return arbitrarily large payloads that the older path would reject. Please enforce the same size limit here before returning the extracted result.

ILindsley:
Re-review against `df85d3079ee97e81295887662f72b1cf8ec31dae` found 2 remaining findings.

- Critical: 0
- High: 0
- Medium: 2
- Low: 0

Summary:
- `ProgressCallback` in `src/types/engine.ts` no longer matches the actual callback payload shape, so typed consumers will miss the new attachment/audio fields and the timestamp type is wrong.
- The new `streamable-http` MCP path drops the existing `maxResponseSizeBytes` guardrails, which is a transport-parity regression for oversized tool responses.

src/types/engine.ts
ILindsley, line 125: Medium: this file now expands the exported response/history artifact contract, but the exported `ProgressCallback` type later in the same module was not updated to match the callback implementation. `src/services/progress/callback.ts` now sends an ISO-string `timestamp` plus optional `message`, `error`, audio fields, and `attachments`, while `ProgressCallback` still says `timestamp: number` and omits those fields. Since `engine.ts` is the shared API surface for callback consumers, typed clients will model the final callback payload incorrectly unless that exported contract is updated too.

ILindsley:
Re-review against `7b725424a4e3951b0133caff259f91ed29ad27f4`:

The two previously reported medium findings are fixed. I re-checked the updated callback contract and the streamable-http size-limit path, and I did not find any new issues in the follow-up changes.

Good for merge.

PR #177, opened 2026-04-30
fix(orchestrator,sandbox): inject prepare_usfm_source; diagnose silent orchestration deaths

IanLindsley:
## Summary

Two bugs surfaced by a WhatsApp QA conversation today (2026-04-30, ~16:11–16:59 UTC on staging — request IDs `e2385c7d`, `4c354c50`, `08f21435`, etc.). Logs analyzed via cf-logs skill.

### Bug A — `prepare_usfm_source` not callable from `execute_code` sandbox

It is an **internal tool** dispatched in the orchestrator; only MCP catalog tools are injected as host functions in the QuickJS sandbox. Claude tried `await prepare_usfm_source(...)` inside `execute_code` three times in the same conversation, hit ReferenceError-equivalents, called `list_tools()` to find correct names — but `list_tools` only returns MCP catalog tools — and concluded "tools aren't in the active registry."

**Fix:** add a curated `SANDBOX_INTERNAL_TOOLS` list (just `prepare_usfm_source` for now). Inject those alongside MCP tools and dispatch via the existing macro-tool handlers. Other internal tools intentionally stay out (see doc comment for `request_audio`, `read_memory`, etc. rationale).

### Bug B — Orchestration silently terminated after iteration 4 (twice)

Two separate turns: `prepare_usfm_source` completed (top-level worked), next `claude_request` fired with the new tool result in context, then **no `claude_response`, no `orchestration_summary`, no `process_chat_complete`**. Zero error/warn-level worker logs. Loop just stopped emitting events.

No proven root cause yet — this PR is the **diagnostic instrumentation** that will make the next recurrence loud:

- **Iteration boundaries**: `iteration_start` / `iteration_end` with reason (`end_turn`/`no_tool_calls`/`continuing`) and duration.
- **Loop boundaries**: `orchestration_loop_start` / `orchestration_loop_end` with `exit_reason` (`done`/`max_iterations`) and `last_iteration`.
- **Anthropic fetch**: `claude_fetch_start`/`_response_received`/`_complete` with body size, status, stop_reason, content blocks, durations.
- **SSE stream**: `sse_stream_start`/`sse_stream_complete` with chunks, bytes, per-event-type counts, elapsed_ms. `sse_stream_no_message` at ERROR-level if stream ends empty.
- **Per-iteration try/catch** around `callClaude` and `executeToolCalls`: `iteration_claude_call_failed`/`iteration_tool_execution_failed` with iteration index + message_count.
- **Max iterations**: `orchestration_max_iterations_reached` at WARN-level (previously silent).
- **90-second AbortSignal** on the Anthropic fetch — without it a stalled upstream hangs forever. With it: `claude_fetch_aborted` + `sse_stream_aborted` with how far the stream had progressed. This is the most likely root cause of the silent death.

### Out of scope (filed separately)

- **Persist `tool_use`/`tool_result` blocks in chat history.** Bug C from the same investigation: when Claude dropped to the next user turn, history showed only the prior text reply, so Claude said "I haven't started yet" despite having actually run tools. Architectural fix; ~1 day; tracked separately.

## Test plan

- [x] Type-check, lint, full test suite — all 516 tests pass.
- [ ] Smoke on staging:
  - [ ] Send a custom-layout request (e.g. "tighten line spacing on James"). Confirm Claude can call `prepare_usfm_source` from inside `execute_code` without "tool not in registry" failure.
  - [ ] Trigger or wait for a long custom-config flow. Confirm new iteration/sse/fetch events all appear in cf-logs and that any silent death now produces an `iteration_claude_call_failed`, `claude_fetch_aborted`, or `orchestration_max_iterations_reached` log.
  - [ ] Verify standard `generate_scripture_pdf` happy path is unaffected.


IanLindsley:
Reviewed PR #177 focusing on the orchestrator timeout/streaming changes and the new execute_code sandbox tool exposure.

Findings:
- High: 1
- Medium: 0
- Low: 0

Top issue:
1. High: the new SSE timeout path can still return a partial Claude message as a successful response. The abort listener cancels the reader, but `parseSSEStream()` never records that the stream was aborted and never throws once a partial `message_start` has been seen. If the 90s timeout fires after some chunks arrive, the loop can exit with `state.message` populated and return truncated text or a half-built tool call instead of surfacing the timeout loudly.

Inline comment added on the relevant hunk. Review limitation: I only sampled the changed paths locally; I did not run an end-to-end timeout repro against Anthropic.

src/services/claude/orchestrator.ts
```diff
@@ -227,101 +300,250 @@ function buildMessageBody(ctx: OrchestrationContext, stream: boolean): string {
...
+): (() => void) | undefined {
+  if (!signal) return undefined;
+  const onAbort = () => {
+    logger.warn('sse_stream_aborted', {
+      elapsed_ms: Date.now() - ref.start,
+      chunks: ref.chunks,
+      bytes: ref.bytes,
+      event_counts: { ...ref.state.eventCounts },
+    });
+    reader.cancel('upstream-abort').catch(() => {
```
IanLindsley, line 504: High: this timeout path can still return a partial Claude message as success.

Why it matters:
When the 90s abort fires after `message_start` or a few deltas have already arrived, the listener cancels the reader, `reader.read()` can finish with `done=true`, and `parseSSEStream()` will return `state.message` as long as it is non-null. That means the turn can end with truncated assistant text or a half-built tool call instead of surfacing the timeout loudly, which is the failure mode this PR is trying to eliminate.

Suggested direction:
Record that the stream was aborted and throw after the read loop whenever `signal.aborted` (or an `aborted` flag) is set, even if a partial message object exists.

src/services/claude/orchestrator.ts
```diff
@@ -1071,48 +1389,18 @@ function checkServerHealth(toolName: string, serverId: string, ctx: Orchestratio
...
-}
-
 async function handleMCPToolCall(
   toolName: string,
   input: unknown,
   ctx: OrchestrationContext
 ): Promise<unknown> {
   const { server } = validateMCPToolCall(toolName, ctx);
-  checkMCPBudgetBeforeCall(toolName, server.id, ctx);
   checkServerHealth(toolName, server.id, ctx);
```
IanLindsley, line 1474: High: removing the request-wide MCP budget here reopens runaway downstream fan-out.

Why it matters:
`MAX_MCP_CALLS_PER_EXECUTION` only limits host calls inside a single `execute_code` sandbox run. After this change, normal top-level MCP tool calls and repeated iterations are no longer bounded by the old per-request downstream budget, and this PR also raises `MAX_ORCHESTRATION_ITERATIONS` from 10 to 30. A docs-heavy discovery loop can now make far more downstream API calls in one user turn than before, which brings back the exact operational-safety problem the deleted budget code was protecting against: rate-limit pressure, cost spikes, and very long turns that fail late instead of fast.

Suggested direction:
Keep the timeout/streaming fixes, but preserve an equivalent request-wide budget guard in `handleMCPToolCall()` (or replace it with another whole-request cap) before raising the iteration ceiling.

IanLindsley:
Re-reviewed PR #177 focusing on the updated orchestration timeout fix and the request-level guard rails around MCP/tool fan-out.

Current result:
- Previous SSE-timeout partial-message issue: fixed
- High: 1
- Medium: 0
- Low: 0

Remaining issue:
1. High: the PR removes the request-wide MCP downstream budget while also increasing `MAX_ORCHESTRATION_ITERATIONS` from 10 to 30. `MAX_MCP_CALLS_PER_EXECUTION` still limits a single `execute_code` sandbox run, but ordinary top-level MCP tool calls and repeated iterations are no longer capped at the whole-request level. That reopens runaway downstream fan-out, rate-limit pressure, and very long failing turns.

Inline comment added on the relevant orchestrator hunk. Review limitation: I did not run a live staging repro of a high-fan-out conversation; this finding is from the changed control flow and the removed budget enforcement code.

IanLindsley:
Re-reviewed PR #177 at commit `8f193e3f672ca8ebfb1545426a75cabb4c8c1065`.

Result:
- Previous SSE-timeout issue: still fixed
- Previous request-wide MCP budget concern: still unresolved
- High: 1
- Medium: 0
- Low: 0

The current head still removes the whole-request downstream budget (`src/services/mcp/budget.ts`, env vars, and `checkMCPBudgetBeforeCall()`), while keeping `MAX_ORCHESTRATION_ITERATIONS` at 30. I did not find any replacement request-level cap in the active orchestrator path, so the prior High finding remains.

No new issues beyond that were identified in this re-review.

wrangler.toml
```diff
@@ -2,13 +2,34 @@ name = "bt-servant-worker"
...
+# way as a hung fetch (no exception, outcome=ok), which is what
+# tail-worker now flags via long_invocation events. See:
+# https://developers.cloudflare.com/changelog/post/2025-03-25-higher-cpu-limits/
+[limits]
+cpu_ms = 300000
+
 [vars]
 ENVIRONMENT = "development"
-MAX_ORCHESTRATION_ITERATIONS = "10"
+MAX_ORCHESTRATION_ITERATIONS = "100"
```
ILindsley, line 32: High: this baseline still allows unbounded request-level MCP fan-out, and this config now makes the blast radius materially larger.

Why it matters:
The old whole-request downstream budget is still gone from `handleMCPToolCall()`, so ordinary top-level MCP tool calls and repeated iterations are no longer capped across a user turn. This commit then raises `MAX_ORCHESTRATION_ITERATIONS` again from 30 to 100 and extends the worker CPU budget to 5 minutes. That means a docs-heavy or otherwise chatty tool loop can now run far longer and issue far more downstream calls before anything stops it. The body-size cap and per-`execute_code` MCP limit do not protect this path.

Suggested direction:
Before increasing the iteration/CPU ceilings, restore an equivalent request-wide cap in the active MCP call path (or add another whole-request guard with the same effect) so runaway conversations fail fast instead of consuming minutes of CPU and a large number of downstream API calls.

ILindsley:
Reviewed PR #177 from the current head (`2f24bbd075d157dbff5aa715b5e55b8a48208413`) as a fresh baseline, focusing on the latest orchestrator, QuickJS, ptxprint, and tail-worker changes.

Findings:
- High: 1
- Medium: 0
- Low: 0

Top issue:
1. High: the request-wide MCP budget is still removed, while the latest head increases `MAX_ORCHESTRATION_ITERATIONS` to 100 and raises the worker CPU cap to 5 minutes. That combination lets ordinary top-level MCP tool loops run much longer and fan out much farther across downstream APIs before anything stops them. The new request-body cap and per-`execute_code` MCP limit do not protect this path.

Inline comment added on the new `wrangler.toml` hunk because that config bump is what materially changes the risk profile since the last review.

Review limitations: this was a static review of the current diff plus targeted local code inspection. I did not run a live staging repro, and I did not rerun the full test suite.

src/services/claude/orchestrator.ts
```diff
@@ -1071,29 +1439,26 @@ function checkServerHealth(toolName: string, serverId: string, ctx: Orchestratio
...
+ */
+function checkAndCountRequestMCPCall(toolName: string, ctx: OrchestrationContext): void {
+  ctx.mcpCallsMade.count++;
+  if (ctx.mcpCallsMade.count > ctx.maxMcpCallsPerRequest) {
+    ctx.logger.warn('mcp_request_call_limit_exceeded', {
+      tool_name: toolName,
+      calls_made: ctx.mcpCallsMade.count,
+      limit: ctx.maxMcpCallsPerRequest,
     });
+    throw new MCPRequestCallLimitError(ctx.mcpCallsMade.count, ctx.maxMcpCallsPerRequest);
```
ILindsley, line 1464: Medium: this new whole-request limit is thrown, but final-mode callers still receive it as a generic 500.

Why it matters:
`MCPRequestCallLimitError` is modeled as a user-correctable 429-style condition, but `processImmediateFinal()` only special-cases `ValidationError`; every other exception becomes `INTERNAL_ERROR` with HTTP 500. So once this cap trips, JSON chat clients will see a server failure instead of an actionable limit response, which makes retries and client handling wrong and hides that the request simply exceeded the per-turn MCP budget.

Suggested direction:
Plumb `AppError`/`MCPRequestCallLimitError` through the final-response path the same way other structured errors should be surfaced, so this limit returns its intended status/code instead of collapsing to 500.

ILindsley:
Re-reviewed PR #177 at the current head (`51027efe50c715aee2fbc52f29c4128d8c387a7c`).

Findings:
- High: 0
- Medium: 1
- Low: 0

Top issue:
1. Medium: the restored whole-request MCP cap now exists, but when it fires the final JSON chat path still collapses it into a generic `INTERNAL_ERROR` 500 instead of surfacing the structured limit condition. That makes a user-correctable quota stop look like a server failure to API clients.

The earlier request-wide fan-out blocker appears addressed in this head: `MAX_MCP_CALLS_PER_REQUEST` is back in config and the active MCP call path increments/enforces it across the whole turn.

Inline comment added on the new throw site. Review limitation: this was static inspection of the latest diff; I did not run an end-to-end request that trips the new limit.

src/services/claude/orchestrator.ts
```diff
@@ -65,6 +57,84 @@ import {
...
+ * stops yielding chunks and no JS-level exception is ever thrown
+ * (`$workers.outcome` stays "ok" because the response 200'd). Failing
+ * fast here surfaces the runaway as a clear `claude_request_body_too_large`
+ * error instead of an indefinite hang the user has no way to escape.
+ * Pairs with truncateToolResultContent — the per-result cap is the
+ * primary defense; this is the seatbelt for whatever it misses.
+ */
+const MAX_REQUEST_BODY_BYTES = 200_000;
+
+class ClaudeRequestBodyTooLargeError extends Error {
```
ILindsley: Medium: this request-body cap is still surfaced to final-mode clients as a generic 500.

Why it matters:
`assertRequestBodyWithinLimit()` throws a bespoke `ClaudeRequestBodyTooLargeError`, but it does not extend `AppError` and `handleOrchestrationError()` just rethrows it. `processImmediateFinal()` only returns structured error responses for `AppError`, so when this limit trips JSON chat clients still get `INTERNAL_ERROR` 500 even though the message clearly describes a user-correctable "start a new conversation / narrow the request" condition.

Suggested direction:
Make this error an `AppError` (or translate it into one in `handleOrchestrationError()`) so the final JSON path can return a structured non-500 response instead of collapsing it to a server failure.

ILindsley:
Re-reviewed PR #177 at the current head (`b944f2385584df32d4059ec08638940999bcfadf`).

Findings:
- High: 0
- Medium: 1
- Low: 0

Top issue:
1. Medium: the final-mode error plumbing now correctly handles `AppError` cases like `MCPRequestCallLimitError`, but the separate request-body-size guard still throws a plain `ClaudeRequestBodyTooLargeError`. That means JSON chat clients still receive a generic `INTERNAL_ERROR` 500 when this user-correctable limit trips.

The earlier request-wide MCP fan-out blocker still appears resolved in this head: `MAX_MCP_CALLS_PER_REQUEST` is configured and enforced in the active MCP call path, and final-mode AppError handling now surfaces that limit correctly.

Inline comment added on the request-body-size error definition. Review limitation: this was static inspection of the latest diff; I did not run an end-to-end request that exceeds the body-size cap.

ILindsley:
Fresh baseline re-review on the current head (`4d608a43430f42a69ba38339ca185ac49b9422fa`).

Result: no remaining findings in this pass.

What I specifically re-checked:
- the SSE timeout path now fails loudly instead of returning partial streamed messages
- the whole-request MCP cap is restored and enforced across the active orchestrator path
- final-mode JSON responses now preserve structured `AppError` status/code, including the request-body-size guard
- the `prepare_usfm_source` sandbox exposure and PDF-attachment flow still look wired correctly after the follow-up changes

Review limits:
- static re-review of the current diff and surrounding codepaths
- did not run an end-to-end staging repro for timeout/body-limit/fan-out scenarios
- older review comments on this PR include now-fixed issues from previous heads

Issue #72, opened 2026-02-20
Add UserQueue Durable Object to replace Fly.io message broker


IanLindsley:
## Context

The [bt-servant-message-broker](https://github.com/unfoldingWord/bt-servant-message-broker) currently runs on Fly.io with Redis (Upstash) to provide per-user message queuing, FIFO ordering, and SSE stream proxying between clients and the worker. This adds an extra service, an extra hosting provider, and an extra data store to the architecture.

Since the worker already runs on Cloudflare with Durable Objects, we can replace the entire broker with a **new Durable Object** in this repo — eliminating Fly.io, Redis, and the Python service entirely.

**Supersedes:** #71 (Simplify 429 lock mechanism)
**Related:** [bt-servant-web-client#2](https://github.com/unfoldingWord/bt-servant-web-client/issues/2), [bt-servant-whatsapp-gateway#2](https://github.com/unfoldingWord/bt-servant-whatsapp-gateway/issues/2)

---

## Architecture Overview

```
                Web Client (SSE)              WhatsApp Gateway (webhook)
                     │                                  │
                POST /api/v1/message              POST /api/v1/message
                     │                            (with callback_url)
                     │                                  │
                     ▼                                  ▼
┌──────────────────────────────────────────────────────────────────┐
│                     Cloudflare Worker (index.ts)                  │
│                                                                  │
│  Routes /api/v1/message and /api/v1/stream to UserQueue DO       │
│  Routes /api/v1/orgs/... to UserSession DO (unchanged)           │
│  Routes /api/v1/admin/... to admin handlers (unchanged)          │
└──────────────────────────┬───────────────────────────────────────┘
                           │
                  env.USER_QUEUE.get(id).fetch(request)
                           │
                           ▼
┌──────────────────────────────────────────────────────────────────┐
│              UserQueue DO (NEW — one per user)                    │
│                                                                  │
│  Storage:                                                        │
│    'queue'       → [{msg1}, {msg2}, ...]  (pending messages)     │
│    'processing'  → true/false                                    │
│    'responses'   → {msg_id: response}     (fallback storage)     │
│                                                                  │
│  In-Memory:                                                      │
│    streams → Map<message_id, WritableStream> (SSE connections)   │
│                                                                  │
│  fetch() → enqueue messages, serve SSE streams, return status    │
│  alarm() → dequeue next message, forward to UserSession DO       │
└──────────────────────────┬───────────────────────────────────────┘
                           │
[...]

PR #202, opened 2026-05-08
feat(admin): add language CRUD endpoints (#197)

IanLindsley:
## Summary

- Adds the per-org Languages CRUD surface that mirrors the existing modes endpoints (GET list, GET one, PUT, DELETE under `/api/v1/admin/orgs/:org/languages[...]`). The portal's Languages tab (admin-portal #92/#93 already merged) calls these paths today and gets 404s — this is the worker side of the wire.
- Storage at `${org}:languages` in `PROMPT_OVERRIDES` KV, read-modify-write matching the modes pattern. Auth inherits the existing admin-token gate. Per-language `language_rights` enforcement stays upstream in the admin-portal BFF.
- Data-shape asymmetry vs modes: a language is `{ name, label?, document, published? }` — single markdown `document` (cap 64000 chars), no slotted overrides. Limits: 20 languages/org, 64-char name, 100-char label.
- One small refactor: `stripControlChars` is now exported from `prompt-overrides.ts` so `languages.ts` can reuse it on document writes (rather than duplicating the regex).

This is one of the five worker-side cuts of admin-portal Epic #72 (others: #198 scaffold endpoint, #199 inline trigger parser, #200 mode markdown migration, #201 Ulysses comment stripping). Closes #197.

## Test plan

- [x] `pnpm check` clean
- [x] `pnpm lint` clean (0 errors; pre-existing warnings only)
- [x] `pnpm test` — 529 passed (32 files), including new `upsert-language` and `validate-language` suites covering creation, document replacement, control-char stripping, max-count cap, label/published carry-through semantics, and validator branches
- [x] `pnpm build` clean
- [x] `pnpm architecture` — no circular deps
- [ ] Staging deploy validates wire shape end-to-end (`curl` GET empty / PUT / GET / DELETE round-trip)
- [ ] Codex review passes

## Open question for review

`DELETE` on an unknown language returns 404, mirroring the modes DELETE handler (`src/index.ts:594`). Issue #197 AC says "DELETE is idempotent" — strictly idempotent would be 200 on missing. Mirroring modes felt like the right default; happy to flip to 200-on-missing if that's the preferred reading (one-line change, would also raise the question for modes).


src/index.ts
```diff
@@ -610,6 +618,148 @@ app.delete('/api/v1/admin/orgs/:org/modes/:modeName', async (c) => {
...
+    const orgLanguages = (await c.env.PROMPT_OVERRIDES.get<OrgLanguages>(
+      `${org}:languages`,
+      'json'
+    )) ?? {
+      languages: [],
+    };
+
+    const filtered = orgLanguages.languages.filter((l) => l.name !== languageName);
+    if (filtered.length === orgLanguages.languages.length) {
+      return c.json({ error: `Language '${languageName}' not found` }, 404);
```
IanLindsley: Issue #197 explicitly lists `DELETE /languages/{name}` as idempotent. Returning 404 when the language is already absent means a retry or repeated delete changes from success to failure, which breaks that API contract and can surface false errors in the portal. Consider treating the missing case as a successful no-op, returning the current `{ org, languages, message }` shape without writing KV.

IanLindsley (commented):
Found 1 medium issue. I cannot submit REQUEST_CHANGES because the active account owns this PR, but this should be treated as blocking: the new language DELETE endpoint does not satisfy the idempotency contract from issue #197, so repeated deletes can surface as false failures to API consumers.

IanLindsley (commented):
All issues addressed. The DELETE /languages/{name} idempotency finding is fixed in 01f7533; the endpoint now returns success when the language is already absent and preserves the expected response shape. GitHub would not allow an APPROVE review because the active account owns this PR.

PR #205, opened 2026-05-09
feat(admin): add per-org language scaffold endpoint (#198)

IanLindsley:
## Summary

- Adds GET/PUT/DELETE endpoints at `/api/v1/admin/orgs/:org/language-scaffold` for managing a per-org markdown template used to pre-populate new language documents
- GET returns a bundled default scaffold (5 H2 sections with `%%` guidance comments) when no org-level override exists in KV; PUT overrides it per-org; DELETE resets to default
- Storage uses existing `PROMPT_OVERRIDES` KV namespace at key `${org}:language-scaffold`

Closes #198

## Test plan

- [x] Unit tests for `validateLanguageScaffold` (valid/invalid inputs, default scaffold integrity) — 10 tests passing
- [x] Type check (`pnpm check`) passes
- [x] Lint (`pnpm lint`) passes with 0 errors
- [x] Architecture check (`pnpm architecture`) — no circular deps
- [ ] CI passes on Linux (includes e2e tests skipped locally on Windows)


IanLindsley (commented):
Found 1 medium issue. GitHub will not allow this account to request changes on its own PR, so this is submitted as a comment review, but the finding should be treated as blocking before merge.

src/index.ts
```diff
@@ -761,6 +767,81 @@ app.delete('/api/v1/admin/orgs/:org/languages/:languageName', async (c) => {
...
+    logger.log('admin_action', {
+      action: 'get_language_scaffold',
+      org,
+      is_default: stored === null,
+      document_length: scaffold.document.length,
+    });
+    return c.json({ org, scaffold });
+  } catch (error) {
+    logger.error('admin_action', error, { action: 'get_language_scaffold', org });
+    return c.json({
```
IanLindsley: This should behave like the other admin config reads and fail closed on storage errors instead of returning the bundled default. The portal will use this response to pre-populate empty language documents; if KV is temporarily unavailable or the read fails for an org with a custom scaffold, returning `200` with the default silently seeds the wrong template and hides the outage from the admin. Please return a 500 here (as modes/languages do) so callers can retry or surface the problem rather than treating the default as authoritative.

IanLindsley (commented):
All issues addressed. Prior medium finding is fixed by returning 500 on scaffold KV read failure; no new issues found in re-review.

PR #208, opened 2026-05-09
refactor(classifier): deterministic slug matching for #mode @language triggers

IanLindsley:
## Summary

- Replace the Haiku LLM classifier with deterministic case-insensitive slug matching
- Same preprocessing architecture: tokens extracted before system prompt assembly, mode overrides + language document shape THIS turn, stripped message goes to orchestration and history
- No external API call, no latency, no cost per message

## Why

The Haiku classifier (PR #206) added an API call per message to fuzzy-match trigger tokens. This was unnecessary — the trigger syntax uses exact slugs (`#spoken`, `@arabic`) that can be matched deterministically. The preprocessing architecture was correct; only the matching mechanism needed to change.

## What changed

| File | Change |
|------|--------|
| `src/services/classifier/index.ts` | Replaced Haiku API call with `extractLeadingTokens()` + `matchSlug()` — pure string matching |
| `src/durable-objects/user-do.ts` | Simplified `classifyAndResolveTriggers()` — no more apiKey/logger in context, synchronous call |
| `tests/unit/classifier.test.ts` | 18 deterministic tests replacing 11 mock-based tests — no fetch mocking needed |

## How it works

1. `extractLeadingTokens()` scans whitespace-separated words at message head that start with `#` or `@`
2. `matchSlug()` does case-insensitive comparison against available mode/language slugs
3. Matched tokens resolve mode overrides and language document into system prompt for THIS turn
4. Unmatched tokens produce soft warnings prepended to the response
5. Stripped message (without trigger tokens) goes to orchestration and history

## Test plan

- [x] `pnpm check` — TypeScript passes
- [x] `pnpm lint` — 0 errors
- [x] `pnpm test` — 542 tests pass (32 files)
- [x] `pnpm architecture` — no circular dependencies
- [ ] CI passes
- [ ] Manual: `#spoken @arabic How do I translate Genesis 1:1?` → response shaped by spoken mode + Arabic language doc, history stores only the question

PR #209, opened 2026-05-09
feat(classifier): restore Haiku LLM classifier for fuzzy trigger matching

IanLindsley:
## Summary

Restore the Haiku LLM classifier for `#<mode> @<language>` trigger token extraction, replacing the deterministic slug matcher from PR #208. This meets the issue #199 acceptance criterion requiring an LLM-driven prompt pattern with forgiving/fuzzy matching.

## What changed

The classifier internals in `src/services/classifier/index.ts` are swapped back to the Haiku API call. The integration in `user-do.ts` is updated to pass `apiKey` and `logger` to the classifier context. Everything else (preprocessing architecture, system prompt injection, token stripping, warning prepend) is unchanged.

## How it works

1. Pre-filter: skip LLM call if no `#`/`@` in first 100 chars (~95% of messages)
2. Haiku call: fuzzy-matches tokens against available modes/languages
3. Validate: confirm matched names exist in available lists
4. Report raw tokens for unmatched triggers (soft warnings)
5. Graceful degradation on any API failure

## Test plan

- [x] `pnpm check` — passes
- [x] `pnpm lint` — 0 errors
- [x] `pnpm test` — 537 tests pass
- [x] `pnpm architecture` — clean
- [ ] CI passes
- [ ] Manual: `#dbs-coach @arabic Hello` → Arabic DBS response, stripped history


src/services/classifier/index.ts
```diff
@@ -16,115 +18,282 @@ export interface ClassifierResult {
...
- * A trigger token parsed from the message head.
- * `prefix` is '#' or '@', `slug` is the word after the prefix.
- */
-interface ParsedToken {
-  prefix: '#' | '@';
-  slug: string;
-  /** Full raw text including prefix, e.g. "#spoken" */
-  raw: string;
-}
+const CLASSIFIER_MODEL = 'claude-haiku-3-5-latest';
```
IanLindsley: High: this model ID is not a valid Anthropic Haiku 3.5 identifier. Anthropic documents the Haiku 3.5 API model as `claude-3-5-haiku-20241022`, with alias `claude-3-5-haiku-latest`; `claude-haiku-3-5-latest` will be rejected, so every trigger message falls into the HTTP-error fallback and the restored classifier never applies the requested mode/language. See https://docs.anthropic.com/en/docs/models-overview.

src/services/classifier/index.ts
```diff
@@ -16,115 +18,282 @@ export interface ClassifierResult {
...
+      });
+      return fallbackResult(messageText, latencyMs);
+    }
+
+    const { modeName, languageName, warnings } = validateClassifierMatches(parsed, ctx);
+
+    return {
+      modeName,
+      languageName,
+      strippedMessage: parsed.stripped_message || messageText,
```
IanLindsley: Medium: this trusts the classifier model to rewrite the user message and then sends/saves that rewritten text as canonical history. If the model hallucinates, is steered by prompt text, or simply over-strips, the user's actual request can be silently changed or dropped. The LLM should only identify the raw leading tokens; derive the stripped message locally, or at least verify `stripped_message` is exactly the original message minus recognized leading triggers before accepting it.

IanLindsley (commented):
Found 1 high and 1 medium issue. I would request changes, but GitHub does not allow this account to request changes on its own PR. The classifier currently calls an invalid Anthropic model ID, so the restored trigger flow will never classify successfully. There is also a message integrity risk from trusting the model-provided rewritten user text.

IanLindsley (commented):
All issues addressed. Re-review verified the model ID fix and the local token stripping change; no remaining findings.

PR #212, opened 2026-05-11
feat(classifier): deterministic cascade + LLM-driven disambiguation (#211)

IanLindsley:
## Summary

- Rewrites `src/services/classifier/index.ts` as a synchronous deterministic parser using a three-tier cascade: exact case-insensitive → unique prefix winner → unique Levenshtein ≤ 2 strict winner.
- Repositions the LLM downstream of parsing: tokens that fail the cascade are surfaced to the orchestrator as structured `unmatchedTriggers` context, and Opus composes a contextual "did you mean…" reply through a new `## Unrecognized Triggers` system-prompt section.
- Deletes `prependTriggerWarnings` and the mechanical underscored warning UX. Drops the Haiku classifier API call, its 600–1500 ms latency, and all the markdown/JSON parse machinery.
- Telemetry: `trigger_classifier_result` drops `classifier_ran` / `classifier_latency_ms` / `warnings`, adds `unmatched_count` and `unmatched_kinds`.

Closes #199. Supersedes #210 — close that PR after this one merges.

## Why

Staging curl tests against the LLM classifier (after #210's tool-use refactor) still showed unacceptable non-determinism for what is fundamentally string matching against a 3–10 element slug list: reverse-order valid tokens (`@english #fia-coach`) flagged as unknown, a clean `@spanish` returning the literal string `"null"`, plus 600–1500 ms latency on the critical path. The cascade gives forgiving fuzzy matching (`#mast` → `mast-methodology`, `#spokne` → `spoken`, `@enzish` → `english`) in microseconds, deterministically.

## Test plan

- [x] `pnpm check` — clean
- [x] `pnpm lint` — clean (0 errors)
- [x] `pnpm test` — 572/572 passing, includes the rewritten `tests/unit/classifier.test.ts` (deterministic cascade cases) and new `tests/unit/system-prompt.test.ts` directive-injection cases
- [x] `pnpm build` — bundles
- [x] `pnpm architecture` — no new circular deps
- [ ] After auto-deploy to staging, curl the acceptance cases against `bt-servant-worker-staging`:
  - `@english #fia-coach hello` → both resolved (reverse order)
  - `@spanish hola` → resolves spanish; no `'@null'` warning
  - `@enzish #fza-ocahch hello there` → `enzish` fuzzy-matches to `english` (distance 2 unique), `fza-ocahch` is unmatched, orchestrator composes a contextual reply through the new `## Unrecognized Triggers` directive
  - `@klingon hi` → unmatched, orchestrator names available languages and answers naturally
- [ ] Verify `trigger_classifier_result` log event in CF Workers logs uses the new fields


IanLindsley (commented):
Review result: no actionable findings from the static pass.

I checked the deterministic classifier rewrite, the `UserDO` trigger handoff, unmatched-trigger prompt injection, and the updated classifier/system-prompt tests. I did not find correctness, security, or maintainability issues worth blocking.

Verification gap: I could not run the focused tests locally because this shell is on Node v20.20.2 and the repo requires Node >=22; Vitest fails during Miniflare/undici startup before executing project tests.

src/services/classifier/index.ts
```diff
@@ -1,330 +1,239 @@
...
-function hasTriggerPrefix(message: string): boolean {
-  const prefix = message.slice(0, PRE_FILTER_PREFIX_LENGTH);
-  return prefix.includes('#') || prefix.includes('@');
-}
+    // `#` or `@` alone (no following word chars) is not a trigger — bail out
+    // without consuming it, leaving it in the stripped message.
+    if (raw.length === 0) break;
 
-// ─── LLM response parsing ──────────────────────────────────────────────────
+    tokens.push({ sigil: ch, raw });
```
IanLindsley, line 86: This still consumes every leading `#...` or `@...` token before we know whether it is a real routing directive. The latest prompt change explicitly treats examples like `@gmail.com`, `#hashtag`, `@team`, and `#1` as usually incidental, but by this point those tokens have already been removed from `strippedMessage`. With input such as `@gmail.com is my address -- can you use that?` or `#1 priority is John 3:16`, Claude receives only the remainder as the user message and is now told to stay silent/proceed normally, so meaningful user content can disappear. The classifier should avoid stripping unmatched tokens that look incidental, or reinject/preserve them in the user message when the directive says to treat them as normal content.

IanLindsley (commented):
Found 1 medium issue on re-review after the latest commit. I cannot submit REQUEST_CHANGES because GitHub rejects change requests from the PR author account, but this should be treated as blocking until fixed.

IanLindsley (commented):
All issues addressed.

Re-reviewed commit `6b8cd85`. The prior medium finding is fixed: unmatched leading `#...` / `@...` tokens are preserved in `strippedMessage`, while matched routing tokens are still removed. The added tests cover the gmail, hashtag, list-marker, mixed matched/unmatched, and case-preservation cases.

Verification run with Node v22.22.2:
- `pnpm check` passed
- `pnpm test tests/unit/classifier.test.ts tests/unit/system-prompt.test.ts` passed: 68 tests
- `pnpm test` passed: 577 tests

Issue #77, opened 2026-03-03
UserQueue DO hits 'Too many subrequests' during polling


IanLindsley:
## Problem

When the web client polls for results via `/api/v1/chat/queue/poll`, it intermittently receives a "Too many subrequests" error from Cloudflare. The error surfaces in the web client as:

```
Polling error: Too many subrequests.
```

This was observed after the UserQueue DO was deployed in PR #73.

## Error Path

1. Web client (`bt-servant-web-client`) enqueues a message via `POST /api/v1/chat/queue`
2. Web client polls in a loop (every 300ms–1s) via `GET /api/v1/chat/queue/poll`
3. At some point during processing, a poll response returns "Too many subrequests"
4. The web client's catch handler at `src/app/api/chat/stream/route.ts:170` wraps it as `Polling error: Too many subrequests`

## Root Cause (Hypothesis)

Durable Objects are single-threaded. When the UserQueue alarm fires, it:

1. Calls `stub.fetch()` to UserSession DO for orchestration
2. UserSession makes many subrequests: MCP tool discovery (N servers), Claude API calls, MCP tool execution (multiple per iteration × multiple iterations), and optionally progress webhook callbacks
3. While the alarm is running, incoming poll requests to the same DO queue up
4. The alarm's subrequest consumption may cause Cloudflare to reject concurrent requests to the DO with "Too many subrequests"

**Note:** The poll handler itself (`handlePoll`) is lightweight — it only reads from DO storage, no outbound fetches. The issue is that it shares an execution context with the alarm.

**Important caveat:** We haven't confirmed via Cloudflare logs that this is definitively a DO subrequest limit issue. The exact Cloudflare error needs to be verified — it could also be a different layer (e.g., the worker entry point, not the DO itself). The string "Too many subrequests" doesn't appear anywhere in our codebase, so it's a Cloudflare runtime error.

## Relevant Code

- **UserQueue alarm:** `src/durable-objects/user-queue.ts:176-227` — dequeues and processes via UserSession
- **Poll handler:** `src/durable-objects/user-queue.ts:286-339` — reads from DO storage (lightweight)
- **Worker route → DO:** `src/index.ts:73` → `handleQueuePoll()` → `stub.fetch()` to DO
- **Web client polling loop:** `bt-servant-web-client/src/app/api/chat/stream/route.ts:122-160`
- **Orchestration subrequests:** `src/services/claude/orchestrator.ts` — Claude API + MCP tool calls in a loop

[...]

PR #214, opened 2026-05-11
feat(prompts): strip Ulysses-style comments before system prompt assembly (#201)

IanLindsley:
## Summary

Strip Ulysses-style editor comments (`%%` line, `++ … ++` span) from author-supplied content just before it reaches Claude. Companion to admin-portal [#72](https://github.com/unfoldingWord/bt-servant-admin-portal/issues/72) (epic) and [#77](https://github.com/unfoldingWord/bt-servant-admin-portal/issues/77) (portal-side parser).

- New `src/utils/ulysses-comments.ts:stripUlyssesComments(text)` — span-first then line pass; lone `++` runs are left literal and surfaced via `hadUnbalancedSpan`; fast-path short-circuit when neither delimiter appears.
- `src/services/claude/orchestrator.ts` strips every resolved prompt slot and the per-turn language document right before `buildSystemPrompt`. Unbalanced spans emit a `ulysses_unbalanced_span` warning with `prompt_slot:<name>` or `language_document` so authors can find malformed input.
- Stored documents are never mutated — the helper operates on in-memory copies for one turn only.

Closes #201.

## Resolved design decisions

- **`%%` is line-terminated**, not paragraph-terminated, for parity with the admin portal's v1 parser (`bt-servant-admin-portal:src/lib/markdown-headings.ts`).
- **Span pass runs first**, line pass runs on the span-stripped result. Otherwise a multi-line `++ … ++` span whose body contains `%%` would have its closing `++` orphaned.
- **Memory is NOT stripped** — `%%`/`++` are an author convention; user-generated memory content stays unfiltered.
- **`buildSystemPrompt` stays pure** — stripping happens at the orchestrator boundary so existing `system-prompt.test.ts` is unaffected.

## Test plan

- [x] `pnpm check` — clean
- [x] `pnpm lint` — clean (0 errors)
- [x] `pnpm test` — 633/633 passing (+19 new in `ulysses-comments.test.ts`)
- [x] `pnpm architecture` — no circular deps
- [x] `pnpm build` — bundles
- [ ] After staging deploys: PUT a `test-mode` document containing `## Identity\n\nReal content. %% editor note\nSecond line ++inline++ surviving.\n` and chat with `client_id: admin-portal` + `#test-mode`. Verify "Real content." and "Second line surviving." reach Opus but "editor note" and "inline" do not.
- [ ] After staging deploys: repeat with an unbalanced fixture (`++note` without closing) and verify the `ulysses_unbalanced_span` warning fires in CF logs.
- [ ] Restore `test-mode` to original state at end.


IanLindsley (commented):
Review pass complete. I found no correctness, reliability, security, or maintainability issues. GitHub will not let this account approve its own PR, so recording this as a clean comment review instead. Verified locally under Node 22.22.2 with pnpm check, pnpm lint, focused vitest coverage for the new stripper/orchestrator path, and the full pnpm test suite.

PR #221, opened 2026-05-12
fix(spoken-mode): resolve remaining smoke test partial-pass items

IanLindsley:
## Summary

Fixes #220 — resolves the 4 partial-pass scenarios from the spoken-mode smoke test (#217).

- **Scenario B (slow activation):** Rewrote Identity section as imperative directives so Claude activates spoken-mode immediately on first message instead of taking 3-4 turns
- **Scenario E (ambient chatter leak):** Short-circuit in `processChat()` when `addressed_to_bot=false` + non-audio — archives message to history but returns empty response without calling Claude. Removes the now-dead `pushAddressedStatusSection` from the system prompt pipeline
- **Scenario J (re-listening uses TTS):** Added `attach_audio` vs TTS priority guidance to Tool Guidance section so Claude serves stored R2 recordings instead of narrating
- **Scenario K (draft fields sparse):** Added atomic draft acceptance mandate to Memory Instructions requiring all draft fields in the same `update_memory` call

## Test plan

- [x] `pnpm check` — TypeScript type-check passes
- [x] `pnpm lint` — no new lint violations
- [x] `pnpm test` — all 620 unit tests pass
- [x] `pnpm build` — build succeeds
- [x] Pre-commit hooks pass (lint-staged, architecture check)
- [ ] CI passes
- [ ] Codex code review
- [ ] Post-deploy: full 15-scenario smoke test (#217) with zero partial-pass results


src/durable-objects/user-do.ts
```diff
@@ -1448,6 +1451,44 @@ export class UserDO {
...
+   * without ever calling the LLM.  Audio with addressed_to_bot=false still
+   * flows through — spoken-mode treats ambient voice during Step 0 as story
+   * submissions.  Returns `null` when the message should proceed normally.
+   */
+  private async maybeShortCircuitAmbient(
+    body: ChatRequest,
+    loaded: Awaited<ReturnType<UserDO['loadChatContext']>>,
+    logger: RequestLogger
+  ): Promise<ChatResponse | null> {
+    if (body.addressed_to_bot !== false || body.message_type === 'audio') return null;
```
ILindsley, line 1468: This lets every `addressed_to_bot=false` audio message continue into orchestration, but the PR also removed `addressedToBot` from the prompt options. After this point Claude can no longer distinguish an ambient voice note from one addressed to the bot, so ambient audio in later spoken-mode phases, or in any other mode, can produce a normal bot response. The previous prompt rule only treated non-addressed voice as a story submission during Step 0. Please preserve the addressed-status signal for audio turns, or otherwise gate this pass-through to the spoken-mode Step 0 story-collection case.

ILindsley (changes requested):
Found 1 medium issue and 1 low issue. Requesting changes so ambient audio keeps enough context to avoid bot replies outside the intended story-collection case, and so the spoken-mode source document stays aligned with the assembled prompt.

ILindsley (approved):
All issues addressed.

src/durable-objects/user-do.ts
ILindsley: Resolved in 7071d863: audio turns now preserve `addressedToBot` into the system prompt, and spoken-mode explicitly gates ambient voice to Step 0 story collection only.

IanLindsley (commented):
All issues addressed.

PR #224, opened 2026-05-13
fix: persist active mode when set via hashtag (#223)

IanLindsley:
Closes #223.

## Summary
- Treat `#mode-name` hashtag activation as a persisted selection, not just a per-turn override. SELECTED_MODE_KEY is now written when the classifier resolves a different mode than the prior persisted one.
- Recognise reserved clear-intent hashtags (`#default`, `#none`, `#clear`) symmetrically: classifier flags `clearMode`, `applyTriggerOverrides` deletes SELECTED_MODE_KEY and recomputes the per-turn resolved prompts without mode overrides so the current turn answers in default.
- Applies to all clients (admin DOs are keyed separately, so no bleed-over).
- Pure decision logic in `decideModePersistence` (unit-tested in isolation). Two new structured log lines — `mode_persisted_from_hashtag`, `mode_cleared_from_hashtag` — make the persistence path observable.

## Evidence (staging, 2026-05-13)
Worker request `614858a5-c363-4414-857f-20224527e55d` matched `#spoken-mode` and produced the welcome turn, but `mcp_calls_made: 0`, no `set_mode`, no `update_memory`, and `memory_found_empty`. SELECTED_MODE_KEY was never written. Subsequent spoken-mode replies were load-bearing on in-DO chat history alone — exactly the fragility this PR fixes.

## Test plan
- [x] `pnpm test` — 656/656 pass
- [x] `pnpm lint` — 0 errors (22 pre-existing warnings unchanged)
- [x] Unit: classifier recognises `#default`/`#none`/`#clear` (case-insensitive, head-of-message scope only, shadows a published "default" mode)
- [x] Unit: `decideModePersistence` (put / skip-when-equal / delete / no-op / clearMode-wins-over-stray-mode)
- [ ] Staging smoke after deploy:
  1. `@bot #spoken-mode` → spoken-mode welcome
  2. `@bot what mode am I in?` (no hashtag) → still spoken-mode (proves persistence)
  3. `@bot #default` → default-mode reply
  4. `@bot what mode am I in?` → default reply (proves clear persisted)
- [ ] Log verification: `mode_persisted_from_hashtag` / `mode_cleared_from_hashtag` events emit with `prior_mode` / `new_mode` populated

## Out of scope
- TTS-skipped-on-spoken-mode (`audio_flow_tts_skipped: true` observed on every spoken-mode turn) — file a follow-up if not already tracked.
- Ambient handling: bare `#mode` without @-mention in groups still short-circuits and never reaches the classifier — existing/correct behavior.


src/durable-objects/user-do.ts
```diff
@@ -145,6 +145,29 @@ const AUDIO_FORMAT_MIME_MAP: Readonly<Record<string, string>> = Object.freeze({
...
+ * Decide whether a classifier turn should persist or clear the user's selected
+ * mode in DO storage. Pure: no I/O, no logging — caller dispatches the action.
+ */
+export function decideModePersistence(
+  classified: { clearMode: boolean },
+  priorActiveModeName: string | undefined,
+  newEffectiveModeName: string | undefined
+): ModePersistenceAction {
+  if (classified.clearMode) {
+    return priorActiveModeName ? { kind: 'delete' } : { kind: 'none' };
```
IanLindsley, line 163: This makes `clearMode` a no-op when there is no persisted prior mode, even if the classifier also resolved a mode token from the same head-of-message token run. In that case `applyTriggerOverrides` may already have set `activeModeName` and `resolved` from `newEffectiveModeName`, so a message like `#default #spoken hi` with no prior mode runs this turn in `spoken` even though clear intent is supposed to win. Consider having clear intent force default resolution regardless of whether storage had anything to delete, and add a test for clear + mode with no prior selection.

IanLindsley (commented):
Found 1 medium issue. I would request changes, but GitHub does not allow the active account to request changes on its own PR.

IanLindsley (commented):
Re-review complete. The prior medium finding is fixed and I found no new issues. All issues addressed.

PR #226, opened 2026-05-13
fix(classifier): tolerate leading punctuation between trigger tokens (#225)

IanLindsley:
Closes #225.

## Summary
- `extractLeadingTokens` now strips a single `,` / `:` / `;` (plus any following whitespace) between consecutive head-of-message trigger tokens. Period is intentionally not in the set (preserves `@user.name` handling).
- One regex replacement in the existing tokeniser. No changes to `ClassifierResult` shape, the matching cascade, or any caller.

## Why
Telegram autocomplete inserts a comma after a bot mention, so users type `@bt_servant_qa_bot , #spoken-mode`. Today the classifier bails at the comma, never sees `#spoken-mode`, and Claude ends up calling `switch_mode` itself (producing the generic "takes effect on your next message" reply instead of the spoken-mode persona welcome).

## Evidence (staging, 2026-05-13 — pre-fix)
Worker request `2daefb25-b5c7-47d4-b33a-ff25170d81af` at 16:35:40Z (group `-5159313139`):
- `trigger_classifier_result`: no `effective_mode`, no `requested_mode`, only an unmatched `language` kind from the bot mention
- `prompt_overrides_applied`: `mode_overrides: 0`, no `active_mode` field
- Claude tools: `list_modes` → `switch_mode("spoken-mode")` → end_turn
- Reply: "Welcome, Ian! 👋 I've switched you to **Spoken Mode** … takes effect on your **next message**…"

Next turn (`59dee7a4-...` at 16:36:57Z) ran with `active_mode: "spoken-mode"` and `mode_overrides: 7` because Claude's prior `switch_mode` had persisted the mode — workaround, not the intended path.

## Test plan
- [x] `pnpm test` — 664/664 pass (up 7 from main)
- [x] `pnpm lint` — 0 errors (22 pre-existing warnings unchanged)
- [x] Unit: parametrised over `,` / `:` / `;` separator; bot+mode together; bot+mode+language; no-whitespace-after-separator; double-separator bails (negative); period control case (regression-protect against email-like patterns)
- [ ] Staging smoke after deploy: in QA Telegram group, send `@bt_servant_qa_bot , #spoken-mode` (with the comma). Expect the spoken-mode persona's first-turn welcome that asks about leaders, NOT the generic `switch_mode` tool reply.
- [ ] Log verification: `trigger_classifier_result` shows `effective_mode: "spoken-mode"`; `prompt_overrides_applied` shows `mode_overrides: 7` and `active_mode: "spoken-mode"`; `mode_persisted_from_hashtag` event fires.

## Out of scope
- Mid-message hashtag scanning (head-of-message scope is intentional per #199 / #211).
- Recognising bot mentions specifically (the `@bt_servant_qa_bot` token still appears in `unmatchedTriggers` as a language — benign, file separately if we want it stripped).
- The `/mode <name>` Telegram slash command (companion issue [unfoldingWord/bt-servant-telegram-gateway#5](https://github.com/unfoldingWord/bt-servant-telegram-gateway/issues/5), independent fix).


src/services/classifier/index.ts
```diff
@@ -97,6 +103,7 @@ function extractLeadingTokens(message: string): { tokens: ParsedToken[]; strippe
 
     tokens.push({ sigil: ch, raw });
     remaining = spaceIdx === -1 ? '' : remaining.slice(spaceIdx).trimStart();
+    remaining = remaining.replace(/^[,;:]\s*/, '');
```
IanLindsley: This strips `,` / `:` / `;` after every consumed leading token before checking whether another trigger actually follows. That widens the change beyond punctuation *between consecutive tokens*: `@bt_servant_qa_bot , please help` now produces `strippedMessage === '@bt_servant_qa_bot please help'`, silently deleting the comma even though the next token is ordinary message text. The existing classifier policy is careful to preserve unmatched leading content, so the separator should only be consumed when the following non-space character is `#` or `@` (and ideally covered by a negative test for separator + non-trigger text).

IanLindsley (commented):
Found 1 medium issue. I would request changes, but GitHub does not allow the active account to request changes on its own PR.

IanLindsley (commented):
Re-review complete. The prior medium finding is fixed and I found no new issues. All issues addressed.

PR #228, opened 2026-05-13
fix(audio): normalize audio_format so MIME-form callers (gateway) work (#227)

IanLindsley:
Closes #227.

## Summary
- New `normalizeAudioFormat(format)` in `src/services/audio/types.ts` strips an `audio/` prefix and returns the bare `AudioFormat`, or `null` if unsupported.
- `validateAudioInput` (`src/services/audio/cloudflare-stt.ts`) uses the helper instead of a raw `.includes()`, so MIME-form `audio/ogg` and bare-form `ogg` both pass validation.
- `audioFormatToMime` (`src/durable-objects/user-do.ts`) normalizes before the map lookup so R2 archival gets the right content-type (`audio/ogg` instead of `application/octet-stream`).

## Why
Voice messages from the Telegram gateway have been failing with HTTP 400 because the gateway sends `audioFormat: "audio/ogg"` (full IANA MIME, perfectly reasonable) but the worker's whitelist contained only bare extensions and the validator did a direct `.includes()` against that list. The error itself was misleading — it listed `ogg` as supported while rejecting `audio/ogg`.

## Evidence (staging, 2026-05-13 — pre-fix)
Two voice attempts from group `-5224926052`, both failed identically:

| Time (UTC) | Worker request_id | audio_format sent | Result |
|---|---|---|---|
| 17:53:59.981 | `ef508e28-f9e1-4a43-9a86-c98b75fed80d` | `audio/ogg` | 400 `AUDIO_TRANSCRIPTION_ERROR` |
| 17:56:45.076 | `b611eaa4-330c-4f1e-bc9b-17e52ba00f29` | `audio/ogg` | 400 `AUDIO_TRANSCRIPTION_ERROR` |

R2 archival succeeded both times (audio arrived intact) but with the wrong content-type. STT was never called — the request was rejected at format validation.

## Test plan
- [x] `pnpm test` — 683/683 pass (+18 new)
- [x] `pnpm lint` — 0 errors
- [x] Unit: `normalizeAudioFormat` covers all six supported formats in both bare and MIME forms, unsupported extensions/MIMEs, non-audio prefixes (`video/ogg`, `application/ogg`), and the empty-string case
- [x] Unit: `transcribeAudio` accepts `audio/ogg` (the exact failing case); rejects `audio/aac` (MIME with unsupported subtype)
- [ ] Staging smoke after deploy: in QA Telegram group, send a voice message (reply-to-bot or with caption mentioning the bot). Expect transcription + reply, not a 400.
- [ ] Log verification: `audio_flow_stt_begin` fires; no `immediate_final_app_error AUDIO_TRANSCRIPTION_ERROR`; `r2_voice_submission_uploaded` records `mime: "audio/ogg"`.

## Out of scope
- Gateway-side changes (independent — worker should accept both forms regardless).
- Adding new audio formats (`aac`, `opus`) to the supported list.
- Changing the STT provider or pipeline.


src/services/audio/types.ts
```diff
@@ -18,6 +18,20 @@ export interface SpeechSynthesisResult {
...
 
+/**
+ * Normalise an inbound `audio_format` value to its bare extension. Accepts
+ * both bare-extension form (`ogg`) and MIME form (`audio/ogg`) so callers
+ * upstream — the Telegram gateway, future SDKs, test harnesses — can use
+ * whichever feels natural. Returns null when the format is not in the
+ * supported list, so callers can throw a single, accurate error.
+ */
+export function normalizeAudioFormat(format: string): AudioFormat | null {
+  const bare = format.startsWith('audio/') ? format.slice('audio/'.length) : format;
```
IanLindsley: This still only handles MIME values whose subtype happens to equal our extension token. For supported formats we already map `mp3` to `audio/mpeg` and `m4a` to `audio/mp4` in `audioFormatToMime`, but this helper rejects those same real MIME values, and the new test explicitly locks `audio/mpeg` out. That leaves MIME-form callers broken for supported formats outside OGG, with the same misleading `Supported: mp3, ...` error this PR is trying to remove. Please normalize through a canonical MIME-to-format map (including `audio/mpeg` -> `mp3` and `audio/mp4` -> `m4a`) and cover those cases in the tests.

IanLindsley (commented):
Found 1 medium issue. I cannot formally request changes because this account is the PR author, but this should be fixed before merge; see inline comment and review summary.

IanLindsley (commented):
Re-review complete: the prior MIME normalization issue is fixed. `normalizeAudioFormat` now maps canonical MIME values like `audio/mpeg` and `audio/mp4`, includes common variants, and the focused unit tests pass locally. All issues addressed.

Issue #104, opened 2026-03-31
Queue path fails: Anthropic SDK returns 403/1003 while direct path works


IanLindsley:
## Symptom

The `/api/v1/chat/queue` endpoint fails to get responses from Claude. The Anthropic SDK returns `403 error code: 1003` in **3 milliseconds** — too fast for a network round-trip, meaning the request is being rejected locally/at the edge before reaching Anthropic's servers.

The `/api/v1/chat/stream` endpoint works perfectly using the exact same code path inside the same UserSession Durable Object.

## Error Details

```
claude_stream_error: "403 error code: 1003"
Stack: Error: 403 error code: 1003
    at APIError.generate (index.js:5785:14)
    at Anthropic.makeStatusError (index.js:10027:21)
    at Anthropic.makeRequest (index.js:10170:24)
```

Error code 1003 is a Cloudflare error meaning "Direct IP Access Not Allowed." It is NOT a standard Anthropic API error.

## Architecture Context

**Direct path (WORKS):**
```
Client → Stateless Worker → UserSession DO → Anthropic SDK → api.anthropic.com ✅
```

**Queue path (BROKEN):**
```
Client → Stateless Worker → UserQueue DO → (alarm) → UserSession DO → Anthropic SDK → api.anthropic.com ❌
```

Both paths call the same function: `UserSession.processStreamingChat()` → `orchestrate()` → `streamClaudeResponse()` → `client.messages.stream()`. Same DO instance, same `this.env.ANTHROPIC_API_KEY` (verified via logging — identical key prefix and length), same SDK version.

## What Was Verified

1. **API key is valid** — `curl` to `api.anthropic.com` returns 200 with a Claude response
2. **Key is identical in both paths** — Added debug logging to `createOrchestrationContext`, both show `sk-ant-api03-kD...` with length 108
3. **Direct path works consistently** — 3/3 tests return 200, 5-6 second response times
4. **Queue path fails consistently** — 3/3 tests fail with 1003 in 3ms
5. **Raw `fetch` works at ALL depths:**
   - Depth 2 (Worker → Anthropic): 200 ✅
   - Depth 3 (Worker → UserQueue DO → Anthropic): 200 ✅
   - Depth 4 (Worker → UserQueue DO → UserSession DO → Anthropic): 200 ✅
6. **Not the alarm** — Modified queue to process inline (in fetch handler, not alarm). Still fails.
[...]

PR #249, opened 2026-06-16
feat(claude): exponential backoff + retry for Anthropic 429/5xx (#248)

IanLindsley:
## Summary

Adds jittered exponential backoff + retry around the Anthropic Messages API call. The worker calls Anthropic via raw `globalThis.fetch` (PR #104 bypassed the SDK because its HTTP layer 1003s from inside our nested Durable Object chain), which means we also lost the SDK's built-in retry/backoff — so every call was a **single attempt** and any `429` surfaced as a user-visible failure. With ~400 users funneling through one org-level Tier-4 key, `429 rate_limit_error` ("Type 2b rate limited") events were climbing.

First surfaced in the admin chat/tester panel: unfoldingWord/bt-servant-admin-portal#208.

## What changed

`fetchAnthropicWithRetry` in `src/services/claude/orchestrator.ts`:

- **Retries** `429, 500, 502, 503, 504, 529` and transient network errors.
- **Does NOT retry** deterministic `400/401/403/404/422`, or our own timeout `AbortError` (already consumed its 90s budget).
- **Full-jitter exponential backoff** (base 1s, ×2, 30s per-delay cap), honoring `Retry-After` (delta-seconds or HTTP-date) when larger than the computed delay.
- **Per-attempt 90s hang-guard** (fresh `AbortSignal` each attempt) + a **180s overall window** — before sleeping a backoff delay, if it would exceed the window we stop and surface the error.
- **Streaming**: retries happen strictly *before* the response body is consumed, so the stream path can only retry pre-first-byte (mid-stream retry is explicitly out of scope).
- **Observability** (per CLAUDE.md — no silent retries): every retry logs `claude_api_retry` (with status, delay, `Retry-After`, `anthropic-ratelimit-*` headers); exhaustion logs `claude_api_retry_exhausted`. Both use the correlated request logger.

Also uses `await` instead of `return <promise>` in the terminal failure chain so rejection handlers attach eagerly — better stack traces and no transient unhandled rejections.

## Testing

- New `tests/unit/orchestrator-retry.test.ts` (11 cases): retry-then-succeed, exhaustion, `Retry-After` honoring, 4xx fail-fast, overall-window guard, streaming pre-first-byte retry, network-error retry, timeout-abort no-retry, plus pure-helper unit tests.
- Full gate green: `pnpm test` (732 tests), `pnpm check`, `pnpm lint`, `pnpm architecture`, `pnpm build`.
- Will validate on staging that real traffic emits `claude_api_retry` and user-visible `claude_error` 429s drop.

## Out of scope (follow-ups)

- Concurrency cap / token-bucket in front of the shared key.
- Tier-4 limit increase / Priority Tier request to Anthropic.
- The malformed `tool_use.input` 400 bug.
- Optional spike: whether `new Anthropic({ fetch: globalThis.fetch })` lets us restore the SDK's own backoff (staging-only repro).

## Closes

Closes #248.

Note: this also resolves unfoldingWord/bt-servant-admin-portal#208, but GitHub can't auto-close across repos — **please close that issue manually when this merges.**


src/services/claude/orchestrator.ts
```diff
@@ -422,21 +505,191 @@ function logFetchAborted(
...
+ * failure was retryable but we ran out of attempts or window), then surface the error.
+ */
+async function exhaustAnthropicRetries(
+  ctx: OrchestrationContext,
+  streaming: boolean,
+  attempt: number,
+  failure: AnthropicAttemptFailure,
+  reason: 'max_attempts' | 'overall_window_exceeded' | null
+): Promise<never> {
+  failure.cleanup();
```
IanLindsley: This clears the per-attempt abort timer before `throwAnthropicHttpError` reads `failure.response.text()` on the terminal path. If Anthropic sends error headers but stalls the body, the retry code can now hang indefinitely while trying to surface the final HTTP error. The retry path keeps the guard active while draining the body, and the previous non-retry code also read the body before cleanup. Please keep the timeout live until the terminal error body has been read, or add a bounded read for this path.

IanLindsley (commented):
Review result: 1 medium issue. I would request changes, but GitHub does not allow the active account to request changes on its own PR. The focused retry tests pass locally, but terminal HTTP-error handling can still hang after clearing the abort guard.

src/services/claude/orchestrator.ts
IanLindsley: Fixed in df50c8e. Moved `failure.cleanup()` into a `finally` that runs after `failAnthropicTerminally`, so the per-attempt 90s abort guard stays live while `throwAnthropicHttpError` reads `response.text()` on the terminal path — a stalled error body now aborts at the timeout instead of hanging. Added a regression test asserting the error body is read before the guard is cleared.

IanLindsley (commented):
Re-review complete. The previous medium finding is addressed by df50c8e: terminal HTTP error handling now keeps the per-attempt abort guard live until `failAnthropicTerminally` finishes, and cleanup runs in `finally`. Focused verification passed: `pnpm vitest run tests/unit/orchestrator-retry.test.ts` (12 tests), plus `git diff --check main...HEAD`. All issues addressed.

PR #307, opened 2026-07-08
fix(catalog): route to Aquifer before training data when Translation Helps misses (#306)

### LINKED ISSUE #306: BTS falls back to training data instead of Aquifer when a resource is missing from tchelps MCP
# Problem

In preparation for upcoming events in Brazil and Colombia, we need top-notch Brazilian Portuguese (and Spanish) support in BTS. Testing on BTS Prod (Tito 1:1) shows the responses themselves are reasonable, but the retrieval fallback chain is wrong: BTS first looks into tchelps MCP, and when the resource does not exist there for a given language/book, it goes straight to training data instead of checking Aquifer. The user had to explicitly force BTS to look at Portuguese resources in Aquifer before it based its response on what's actually available there.

Image

Image

Image

## Expected behavior

- Fallback chain should be: tchelps MCP → Aquifer → training data (clearly labeled) — not tchelps MCP → training data. BTS should query Aquifer for scripture, notes, and biblical terms in Portuguese without needing to be forced, and only supplement with clearly-labeled training knowledge when no resource exists in either source.
- The new Brazilian Portuguese reference text is available to BTS and used as a drop-in replacement for the BSB wherever we would use the BSB in English responses. https://github.com/BibleAquifer/AquiferPortugueseBibleReferenceText/tree/main/por/pdf

## Actual behavior

When a resource is missing from tchelps MCP for a language/book, BTS skips Aquifer and answers from training data; it only consults Aquifer's Portuguese resources when explicitly prompted to

IanLindsley:
## Summary

Fixes #306 — when Translation Helps (tchelps) lacks a resource for a language/book (e.g. **no Portuguese resources**), BTS was jumping straight to training data (or self-translating English notes) instead of checking Aquifer, which does carry Portuguese content. Users had to *force* BTS to look at Aquifer.

Two coordinated changes:

- **Code (this diff):** `generateToolCatalog` now groups MCP tools under a per-server `###` heading (display name from server config, falling back to the raw server id). The catalog previously rendered one flat table with no server attribution, so BTS couldn't tell Translation Helps tools from Aquifer tools. All data was already on `CatalogTool.serverId` / `ToolCatalog.serverMap` — no type/signature changes.
- **Prompt (applied to staging KV, not in this diff):** the org `tool_guidance` slot now encodes the explicit fallback order **Translation Helps → Aquifer → training data (with permission, clearly labeled)**, and removes the old `"If a response is 404 respond with training data"` short-circuit.

## Test plan

- [x] Reproduced on staging via `POST /api/v1/chat` (Portuguese Titus 1:1, fresh user_ids): pre-fix BTS skipped Aquifer and self-translated English `en_tn`, falsely reporting "no Portuguese resources."
- [x] Prompt-only intermediate check on staging: BTS now says "Translation Helps has no Portuguese resources, I'll try Aquifer" and cites Aquifer sources + the Aquifer Portuguese Reference Text.
- [x] Local gates: `check`, `lint` (0 errors), `test` (792 passed), `build`, `architecture` — all green.
- [ ] Full A+B verification runs after this PR's staging deploy (change B live).

## Notes / follow-ups

- **Out of scope:** #306's second bullet (Portuguese Reference Text as a drop-in BSB replacement in English responses) — deferred.
- **Upstream Aquifer MCP gap (separate issue):** Portuguese *notes/terms* localization is not retrievable via the Aquifer `get`/`entity` tools even though `list`/`browse` advertise `por`. Portuguese *scripture* works. This bounds how much of "notes/terms in Portuguese" the routing fix can deliver until fixed upstream.

Closes #306


IanLindsley (commented):
No issues found in this review.

Verified the catalog grouping behavior against the changed implementation and surrounding system-prompt/tool-dispatch call sites. Local checks passed: `npm test -- tests/unit/mcp-catalog.test.ts tests/unit/system-prompt.test.ts`, `npm run check`, and `npm run lint` (warnings only, outside changed files).

PR #318, opened 2026-07-21
fix: make silent error paths observable (close logging gaps)

IanLindsley:
## Why

An audit of the worker's error handling surfaced a set of paths that fail
**silently** — several of which violate the repo's "never silently swallow
errors" rule. This closes them with structured, request-correlated logs so every
failure is observable. **No control-flow changes**: rethrows still rethrow,
responses are unchanged. Once the OTLP facade swap lands later, each of these
logs becomes a trace-correlated OTLP record for free.

## What it closes

| Area | Fix | Files |
|---|---|---|
| Auth | Log 401/403 rejections (reason + path, never the token) | `src/index.ts` |
| Backpressure | Log 429 rate-limit + queue-full rejections | `src/durable-objects/user-do.ts` |
| Untried I/O | Wrap R2 audio + memory store put/get/delete in try/catch (log→rethrow) | `services/audio/r2-storage.ts`, `services/memory/store.ts` |
| Privacy | Stop logging raw model-authored sandbox code (log length only) | `services/code-execution/quickjs-executor.ts` |
| Pre-throw context | Log STT validation failures before throwing | `services/audio/cloudflare-stt.ts` |
| Best-effort catches | Justify the genuinely-intentional ones (error-body reads, idempotent drain/cancel) with comments | `orchestrator.ts`, `ptxprint/{pdf-mirror,usfm-source}.ts` |

## Notes

- `enforceEnqueueRateLimit` was extracted from `handleUnifiedChat` to keep it
  within the max-lines rule after adding the 429 log.
- `usfm-source.ts`'s `buildSourceUrl` was intentionally left untouched — its
  throws are already logged upstream by macro-tool's `prepare_usfm_source_error`
  catch, so it's not a silent swallow.

## Tests

Adds unit tests for the new R2 + memory store failure paths (log + rethrow) and
the R2 miss path.

## Verification

`pnpm check`, `pnpm lint`, `pnpm build` pass; **767 + new tests pass**. Version
bumped to 2.29.3.


src/services/memory/store.ts
```diff
@@ -416,7 +431,13 @@ export class JsonMemoryStore implements UserMemoryStore {
    * Get memory data, auto-migrating from v1 markdown format if needed.
    */
   private async getMemoryData(): Promise<MemoryStorage> {
-    const raw = await this.storage.get<unknown>(MEMORY_STORAGE_KEY);
+    let raw: unknown;
+    try {
+      raw = await this.storage.get<unknown>(MEMORY_STORAGE_KEY);
+    } catch (error) {
+      this.logger.error('memory_read_failed', error, {});
```
IanLindsley, line 438: **Medium:** This catch only covers the storage read, so the v1 migration path still has an unlogged Durable Object write. When `raw` is a string, `getMemoryData()` calls `migrateV1()`, whose `storage.put()` can reject without emitting any `memory_*_failed` event. That leaves one of this store's `put` paths outside the PR's claimed log-then-rethrow coverage. Please log migration write failures as well and add a regression test for that path (or route the writes through a shared logged helper).

src/index.ts
```diff
@@ -234,6 +247,13 @@ app.use('/api/v1/admin/orgs/:org/*', async (c, next) => {
   const authHeader = c.req.header('Authorization');
 
   if (!authHeader?.startsWith('Bearer ')) {
+    createRequestLogger(crypto.randomUUID()).warn('admin_auth_rejected', {
```
IanLindsley: **Low:** This admin-specific rejection branch, and the 403 branch below it, cannot be reached with the current middleware ordering. The earlier `/api/*` middleware returns for a missing bearer or any token other than `ENGINE_API_KEY`; it only calls `next()` for a valid engine key, which this middleware immediately accepts at line 263. Consequently no `admin_auth_rejected` event (including the `no_org_admin_key` distinction) can ever be emitted. Please emit the admin context in the outer middleware or restructure the auth middleware, and cover the routing behavior with a test.

IanLindsley (commented):
Review result: 0 critical, 0 high, 1 medium, 1 low. Blocking verdict: changes requested. GitHub does not permit a REQUEST_CHANGES event on a self-authored PR, so this is submitted as COMMENT. The memory migration write still lacks the new log-then-rethrow coverage. The admin-specific rejection logs are also unreachable and should be corrected or removed.

src/index.ts
IanLindsley: Addressed in 3fc30cea: the unreachable admin-specific log additions were removed; the reachable global middleware remains the single source of `auth_rejected` events for these routes.

src/services/memory/store.ts
IanLindsley: Fixed in 3fc30cea: the migration write now emits `memory_migration_write_failed` and rethrows, with a regression test covering the failing v1 migration path. Verified by the full local suite.

IanLindsley (commented):
Re-review result: 0 critical, 0 high, 0 medium, 0 low. Both prior findings are fixed and the follow-up patch introduces no new issues. All issues addressed. GitHub does not permit approving a self-authored PR, so this clean review is submitted as COMMENT.

PR #319, opened 2026-07-22
chore(infra): M0 - fly.io OTel collector + OpenObserve sink scaffolding

IanLindsley:
## M0 — Telemetry infrastructure (fly.io collector + self-hosted OpenObserve)

First milestone of the worker-side OpenTelemetry buildout (issue #309, epic bt-servant-telemetry#38). **No worker code ships in this PR** — it adds an `infra/` tree of IaC + runbooks only, which Wrangler does not bundle.

### What this delivers
- `infra/otel-collector/` — OpenTelemetry Collector on fly.io: bearer-authed OTLP/HTTP receiver → `memory_limiter` → defense-in-depth redaction → `batch` → OpenObserve exporters (traces/logs/metrics), plus a `debug` exporter for staged bring-up. `Dockerfile` bakes our config into the contrib image; `fly.toml` keeps the pipe warm (`min_machines_running = 1`).
- `infra/openobserve/` — single-binary self-hosted OpenObserve (store + UI) on fly.io with a persistent volume; the single sink, governance in our hands.
- `infra/README.md` — architecture, the "why a collector at all" rationale, and bring-up order.

### Design notes
- **Collector, not worker→sink direct**: the worker only ever knows one endpoint + one secret. Adding a second sink later is a collector-config-only change the worker never sees.
- **Redact twice**: at source (worker, later milestones) and again here (`attributes/redact` drops auth/token/secret keys, hashes `user_id`). No content/PII reaches the sink.
- **Secrets** live in fly (`fly secrets set ...`), never in git. `${env:...}` placeholders only.
- **Pin before real use**: images intentionally default to `:latest` in scaffolding with explicit "do NOT ship :latest" comments; pin a stable tag at deploy.

### DoD for M0
curl one bearer-authed OTLP payload → record visible in the OpenObserve UI. (Requires the user's fly.io account to actually deploy — this PR lands the reviewable IaC.)

### Version
Bumped to **2.29.5** (patch — infra/IaC only, no worker behavior change).

Follow-ups per the approved plan: M1 traces from worker context → M2 logger facade swap → M3 manual spans → M4 bounded metrics.


infra/openobserve/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+```bash
+# from infra/openobserve/
+fly launch --no-deploy                       # edit fly.toml app/region first
+fly volumes create openobserve_data --size 3 # persistent storage for /data
+
+# Root user is created on first boot from these secrets.
+fly secrets set \
+  ZO_ROOT_USER_EMAIL="you@example.com" \
+  ZO_ROOT_USER_PASSWORD="$(openssl rand -hex 24)"
```
IanLindsley: **[medium] Preserve the bootstrap password before setting the secret.** This generates the password only inside command substitution. Fly secrets are write-only, so after the command returns the operator has no password to use for the login on line 21; first boot creates the root account with an unknown credential and the M0 UI verification is blocked. Generate and securely record the password first, then pass that recorded value to `fly secrets set`.

infra/otel-collector/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Deploy
+
+```bash
+# from infra/otel-collector/
+fly launch --no-deploy            # once, to create the app (edit fly.toml app/region first)
+
+# Secrets — the worker→collector shared secret + the OpenObserve credentials.
+fly secrets set \
+  OTEL_INGEST_TOKEN="$(openssl rand -hex 32)" \       # give this same value to the worker
```
IanLindsley: **[medium] Generate and retain the shared token, and fix this continuation.** The generated value is consumed by `fly secrets set` and never retained for the later `wrangler secret put`, so the worker cannot be configured with the same token. Also, the backslash is followed by spaces and an inline comment, so Bash does not continue the command here (it passes a blank argument to `fly` and treats the next two lines as assignments). Generate the token once and securely retain it for both apps; put the comment on its own line and make the backslash the final character.

infra/openobserve/fly.toml
```diff
@@ -0,0 +1,28 @@
+# fly.io app for self-hosted OpenObserve (the telemetry sink: store + web UI in one binary).
+# Adjust `app`/`primary_region` to your fly org before `fly launch`/`fly deploy`.
+app = "bt-servant-openobserve"
+primary_region = "iad"
+
+[build]
+  # Pin to a specific stable tag before real use — do NOT ship :latest.
+  # Tags: https://gallery.ecr.aws/zinclabs/openobserve
+  image = "public.ecr.aws/zinclabs/openobserve:latest"
```
IanLindsley: **[medium] Pin the image in the deployable config.** The runbook executes plain `fly deploy` without a step that replaces this value, so the documented M0 path does ship `:latest` despite the warning above. That makes first deploys and redeploys select unreviewed bits and can silently introduce storage/schema upgrades. Pin a tested OpenObserve version (ideally a digest) here and make upgrades explicit.

IanLindsley (commented):
PR review: 0 critical, 0 high, 3 medium, 0 low. The infrastructure formats and Collector configuration validate, but the documented bootstrap loses both generated credentials, one shell continuation is broken, and OpenObserve is deployed from a mutable image tag. Please treat this as a request-changes verdict and address the inline findings before merge. GitHub does not allow the PR author to submit a formal REQUEST_CHANGES review on their own PR.

IanLindsley (commented):
Re-review of head 29bc4d4148dd9b6f6c7ed3a17a0d8ede58549843: 0 critical, 0 high, 0 medium, 0 low.

All three prior findings are fixed, the updated Fly configurations pass strict validation, the Collector configuration still validates with otelcol-contrib v0.157.0, and current checks are green. GitHub does not allow the PR author to approve their own PR.

All issues addressed.

PR #321, opened 2026-07-22
feat(telemetry): M1 - OTLP traces from the worker context

IanLindsley:
## M1 — OpenTelemetry traces from the worker context

Second milestone of the worker-side OTel buildout (issue #309, epic bt-servant-telemetry#38), building on the M0 collector + OpenObserve infra now live on fly.io. Wires `@microlabs/otel-cf-workers` so the worker emits OTLP traces to our collector.

**Additive and feature-flagged — safe to merge as a no-op.** Telemetry only egresses when *both* `OTEL_EXPORTER_OTLP_ENDPOINT` and `OTEL_COLLECTOR_TOKEN` are set (via `wrangler secret put`). Until then, `resolveTelemetryConfig` returns a valid config with a **0% head sampler**, so no spans are recorded and nothing leaves the worker. The existing `console.log` → Workers Observability path keeps running alone during the parallel bake; rollback is just unsetting the secret.

### What lands
- **`src/services/telemetry/{config,index}.ts`** — `resolveTelemetryConfig` (`ResolveConfigFn`) reads the endpoint + token from `Env`, sets `service.name` / `namespace` (= `ENVIRONMENT`) / `version`, and points the OTLP exporter at `<endpoint>/v1/traces` with a `Bearer` auth header matching the collector's bearer-auth extension. Disabled → 0% sampler + never-contacted placeholder URL.
- **`src/index.ts`** — wrap the Hono app with `instrument()` (request root span + auto-spans on every outbound `fetch`: Anthropic, MCP, TTS, webhooks) and the `UserDO` class with `instrumentDO()` (per-invocation trace context, flushed from the DO fetch context where raw global `fetch` works — sidesteps the CF 1003 alarm-context restriction).
- **`src/config/types.ts`** — optional `OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_COLLECTOR_TOKEN` on `Env`.
- **`wrangler.toml` + `wrangler.test.toml`** — `compatibility_flags = ["nodejs_compat"]` (otel-cf-workers needs AsyncLocalStorage); inherited by the staging env. Kept `compatibility_date` at `2024-01-01` — `pnpm build` bundles cleanly, no node-builtin gaps.
- **`package.json`** — `@microlabs/otel-cf-workers@1.0.0-rc.52` + `@opentelemetry/api@~1.9.0` as production deps, plus a targeted `pnpm.override` pinning `@opentelemetry/core` to `^2.8.0` to clear a moderate transitive advisory (GHSA-8988-4f7v-96qf) and keep `pnpm audit --prod` at zero severity.

### Verification
- `pnpm check`, `pnpm architecture`, `pnpm build` (bundle 655 KiB gzip, +~100 KiB, well within limits), `pnpm test` (**785 pass**, +11 new telemetry tests), `pnpm audit --prod` (zero) all green locally.
- **Runtime no-op check (post-deploy):** because `instrument()` wraps the handler even when disabled, I'll hit staging `/health` after the CI deploy to confirm the instrumented worker still serves with the secrets unset. Then, separately, enable the secrets on staging and confirm a real request produces a span tree in OpenObserve.

### Version
Bumped to **2.30.0** (minor — new feature).

Follow-ups per the approved plan: **M2** logger facade → OTLP logs (+ source-level attribute redaction), **M3** manual spans over non-fetch seams, **M4** bounded metrics.


src/services/telemetry/config.ts
```diff
@@ -0,0 +1,63 @@
...
+      service,
+      // Never contacted: the 0% head sampler below drops every span pre-export.
+      exporter: { url: 'http://localhost:4318/v1/traces' },
+      sampling: { headSampler: { ratio: 0 } },
+    };
+  }
+
+  return {
+    service,
+    exporter: {
```
IanLindsley: [high] This exporter receives sensitive URL attributes before the planned M2 redaction. In rc.52, automatic request/fetch instrumentation adds `url.full`, `url.path`, and `url.query` to spans. Here that includes user/chat/thread identifiers in inbound routes, `user_id` on DO history requests, and potentially signed credentials in caller-supplied webhook or admin-configured MCP query strings. Those values would be retained by the collector/OpenObserve even though this module says the spans are structural. Please redact at the exporter/span-processor boundary before enabling telemetry and add a serialization-level test proving path/query identifiers and secrets never reach OTLP. (`postProcessor` is parsed but not invoked by the pinned rc.52 build, so configuring it alone would not fix this.)

src/services/telemetry/config.ts
```diff
@@ -0,0 +1,63 @@
...
+    namespace: env.ENVIRONMENT,
+    version: APP_VERSION,
+  };
+
+  if (!isTelemetryEnabled(env)) {
+    return {
+      service,
+      // Never contacted: the 0% head sampler below drops every span pre-export.
+      exporter: { url: 'http://localhost:4318/v1/traces' },
+      sampling: { headSampler: { ratio: 0 } },
```
IanLindsley, line 166: [medium] A 0% head sampler does not make rc.52 a no-op. The library still records spans and its default tail sampler exports root errors; because remote parents are accepted by default, a sampled inbound `traceparent` can also override this ratio. Running the suite with both secrets unset reproduced repeated attempts to connect to this localhost exporter. The disabled config should use an always-false tail sampler, reject remote sampling, and disable inbound/outbound trace-context propagation (or bypass instrumentation entirely) so unset secrets cannot egress or generate export failures. Please cover an error response/exception and a sampled inbound `traceparent` in tests.

IanLindsley (commented):
Found 2 blocking issues (1 high, 1 medium): enabled tracing exports sensitive URL identifiers/query values without redaction, and the disabled configuration still attempts exports despite the promised no-op behavior. See the inline comments for evidence and remediation. Independent type-check, architecture, build, and all 810 tests passed, but the test run emitted repeated localhost:4318 connection-refused export errors with telemetry secrets unset, corroborating the disabled-path finding.

src/services/telemetry/config.ts
IanLindsley: Still open on `42d5cd4`. The new tests call `config.postProcessor` directly, but the pinned rc.52 build never invokes that hook: its compiled code only stores `postProcessor` while parsing config, with no call before serialization/export. Actual OTLP spans therefore remain unredacted. There is a second gap as well: Durable Object spans are named `Durable Object Fetch/Alarm ${id.name}` and carry `do.id`/`do.name`, while this repo derives DO names from `user:${org}:${userId}` and `group:${org}:${chatId}:${threadId}`. Even a functioning URL-only processor would leak those identifiers. Please put redaction on the exporter/span-processor path that actually runs, scrub DO names/attributes too, and verify captured serialized/exported spans rather than invoking the helper directly.

src/services/telemetry/config.ts
IanLindsley: Partially fixed on `42d5cd4`. `NOOP_EXPORTER` does eliminate collector/localhost network exports; the full 815-test run no longer emits the connection-refused failures. However, disabled config still leaves rc.52's `handlers.fetch.acceptTraceContext` and `fetch.includeTraceContext` defaults enabled. A sampled inbound `traceparent` is still accepted and a trace header is still injected into every outbound fetch, so telemetry state still leaves the worker with both secrets unset. Please set both controls to `false` on the disabled path (and test a sampled inbound parent plus captured outbound headers) to satisfy the feature-gate/no-op contract.

IanLindsley (commented):
Re-review of `42d5cd4`: both prior blocking findings remain open. The no-op exporter fixes the localhost export attempts, but disabled mode still accepts and propagates trace context. The redaction fix is ineffective because rc.52 never invokes `postProcessor`, and it also misses identifier-bearing Durable Object span names/attributes. Type-check, architecture, build, and all 815 tests pass. Verdict remains request changes; GitHub requires a COMMENT state because the active account authored this PR.

src/services/telemetry/config.ts
IanLindsley: Resolved on `a15f51b`. `RedactingSpanExporter.export()` now invokes redaction immediately before the real OTLP exporter, and the redactor covers URL path/query attributes plus identifier-bearing Durable Object names and attributes.

src/services/telemetry/config.ts
IanLindsley: Resolved on `a15f51b`. Both enabled and disabled configs now set `handlers.fetch.acceptTraceContext` and `fetch.includeTraceContext` to `false`, while the disabled path retains the no-op exporter.

src/services/telemetry/config.ts
```diff
@@ -0,0 +1,170 @@
...
+  const full = toOrigin(attrs['url.full']);
+  if (full !== undefined) attrs['url.full'] = full;
+  const httpUrl = toOrigin(attrs['http.url']);
+  if (httpUrl !== undefined) attrs['http.url'] = httpUrl;
+  delete attrs['url.path'];
+  delete attrs['url.query'];
+  delete attrs['http.target'];
+  delete attrs['do.id'];
+  delete attrs['do.name'];
+  delete attrs['do.id.name'];
```
IanLindsley, line 83: [medium] rc.52 also auto-instruments every KV operation and records the key in both `db.statement` (for example, `get <key>`) and `db.cf.kv.key`. This worker passes the request-derived `org` and values such as `${org}:modes` directly as KV keys, so the organization/tenant identifier still reaches OTLP despite the governance guarantee and URL/DO redaction above. Please drop or redact these KV key/statement attributes at export time and cover a representative KV span in the redaction tests.

IanLindsley (commented):
Re-review of `a15f51b`: the two prior findings are resolved. One new medium privacy finding remains: rc.52's automatic KV spans export request-derived organization keys through `db.statement` and `db.cf.kv.key`, which the exporter redactor does not remove. Type-check, architecture, build, all 818 tests, production audit, and current CI pass. Verdict remains request changes; GitHub requires a COMMENT state because the active account authored this PR.

src/services/telemetry/config.ts
IanLindsley: Resolved on `ddf6e6d`. `redactSpan` now drops both `db.statement` and `db.cf.kv.key` before the wrapped OTLP exporter runs, with a representative KV span test preserving only identifier-free binding/operation fields.

IanLindsley (commented):
Re-review of `ddf6e6d`: the remaining KV tenant-key finding is resolved and no new issues were found. Type-check, architecture, build, all 819 tests, production audit, and current CI pass. The PR is open, non-draft, conflict-free, and GitHub reports no required checks or merge blocker. All issues addressed. GitHub requires a COMMENT state because the active account authored this PR.

PR #322, opened 2026-07-22
fix(mcp): use eval-free JSON schema validator in streamable-http client

IanLindsley:
## Problem

Streamable-HTTP MCP discovery fails with `Code generation from strings disallowed for this context` for any server whose tools declare `outputSchema`. Hit in staging when the `translation-helps` server was repointed to `https://tc-helps.mcp.servant.bible/v2/mcp` — its `list_languages` tool declares an `outputSchema`, the only tool across our registered servers that does (which is why ptxprint-mcp and yaapi-bible were unaffected).

Root cause: the MCP SDK's `Client` defaults to an Ajv-based JSON-schema validator, and Ajv compiles schemas with `new Function` — forbidden in the Workers runtime.

## Fix

- Pass the SDK's own `CfWorkerJsonSchemaValidator` (built for edge runtimes; validates by interpretation, no code generation) to the `Client` in `src/services/mcp/streamable-http-client.ts`
- Add `@cfworker/json-schema` — the SDK's optional peer dependency backing that provider

## Testing

- New regression test `tests/unit/streamable-http-output-schema.test.ts`: mocks a full streamable-HTTP MCP session (initialize → initialized → tools/list) returning an `outputSchema`-bearing tool. The vitest workers pool runs workerd with unsafe-eval enabled, so the test stubs the global `Function` constructor to emulate the production restriction. Verified it fails with the exact production error before the fix and passes after.
- Full suite: 44 files / 820 tests pass; lint, typecheck, architecture, build, and both audit gates clean.
- Live staging verification planned post-deploy: `GET /api/v1/admin/orgs/unfoldingWord/mcp-servers?discover=true` should report `discovery_status: ok` with 13 tools for translation-helps.

## Notes

- Version bumped to 2.30.1 (patch).
- Lockfile refresh from `pnpm add` also bumped `hono` 4.12.25 → 4.12.31 within its existing semver range.


IanLindsley (commented):
Reviewed head 7bc99a4e396392029d72f8387af4b72133c27aef. No critical, high, medium, or low findings. The regression test, full test suite (820 tests), typecheck, formatting check, and Worker build all pass locally. Formal state is COMMENT because GitHub does not allow authors to approve their own pull requests.

PR #323, opened 2026-07-22
feat(telemetry): M2 — tee structured logs to OTLP logs

IanLindsley:
## Summary
- **Facade swap (the #1 M2 goal):** re-back `logger.ts` so every existing `logger.*` call site *also* emits a trace-correlated OpenTelemetry log record, **additively** — the `console.log` → Workers Observability path keeps running during the parallel bake. A genuine no-op until the OTLP endpoint + collector token are set, so rollback is just unsetting the secret.
- **Own logs stack (`src/services/telemetry/logs.ts`):** `@microlabs/otel-cf-workers` is traces-only, and the stock `@opentelemetry/exporter-logs-otlp-http` can't run in Workers (its transports use XHR/sendBeacon/node:http). So M2 ships: a `LoggerProvider`, a **timer-free** `BufferingLogRecordProcessor` (records drain once per invocation at the request/DO boundary via `waitUntil` — the Workers runtime has no reliable background timers, mirroring how otel-cf-workers flushes spans), and a `FetchLogExporter` that serializes with `@opentelemetry/otlp-transformer` (`JsonLogsSerializer`) and POSTs over otel-cf-workers' `__unwrappedFetch` (RAW global fetch, so the export isn't self-traced).
- **Clean wiring:** `logger.ts` stays a pure util and exposes `setLogSink()`; the telemetry layer injects its emitter (dependency direction `services → utils`, no cycle). The worker fetch middleware and `UserDO.fetch` each init + flush independently (separate isolates → separate buffers).
- **Governance:** redact at source (sensitive attribute *keys* masked, nested data JSON-encoded via existing redaction utils) and again in the exporter; identifiers (`request_id`/`user_id`) are kept as correlation keys — `user_id` is pseudonymized (`hash`) by the collector's redact processor, not at source.
- **Deps** pinned to the in-tree `0.200.0` OTel line; dropped the unusable stock log exporter for `otlp-transformer` + `resources`. Both `pnpm audit` gates pass clean.

Advances #309 (worker OTel slice). Sequenced after M1 traces (#321); M3 spans + M4 metrics remain under #309.

## Test plan
- 16 new unit tests in `tests/unit/telemetry/logs.test.ts` (attribute redaction, severity mapping, buffering/single-drain, fetch exporter POST + failure paths, disabled no-op, enabled init→emit→flush→export). **811 total unit tests pass.**
- Local gates all green: `pnpm check` (tsc), `pnpm lint` (0 errors), `pnpm architecture` (no cycles), `pnpm build` (wrangler bundle OK under `nodejs_compat`), `pnpm format:check`, `pnpm audit --prod` + `--audit-level=high`.
- Feature-flagged no-op when telemetry secrets are unset (safe to merge; prod worker has no secrets). Post-merge: enable staging secrets + confirm log records land in OpenObserve correlated by `trace_id`.

## Documented gap
DO streaming/alarm-context late logs flush best-effort (next boundary / tail worker); full DO-lifecycle log coverage lands in M3 with the `withSpan` seams.


src/services/telemetry/logs.ts
```diff
@@ -0,0 +1,293 @@
...
+export function buildLogAttributes(entry: LogEntry): Record<string, AttributeValue> {
+  const attributes: Record<string, AttributeValue> = {};
+  for (const [key, value] of Object.entries(entry)) {
+    if (key === 'event' || key === 'timestamp') continue;
+    if (value === undefined || value === null) continue;
+    if (SENSITIVE_KEY_PATTERN.test(key)) {
+      attributes[key] = '[REDACTED]';
+      continue;
+    }
+    if (typeof value === 'string' || typeof value === 'number' || typeof value === 'boolean') {
```
IanLindsley: **High — Prevent user content and URL credentials from egressing.** This branch forwards every primitive whose key misses `SENSITIVE_KEY_PATTERN`, but the existing facade has reachable calls such as `process_chat_complete` with the full `response`, `stt_complete`/`audio_flow_stt_complete` with transcript previews, and `webhook_failure` with the full callback `url`. The collector redactor only handles credential-named keys plus `user_id`, so these values reach both OpenObserve and the detailed debug exporter. That contradicts the no-message-content/precise-location guarantee and can also expose signed query credentials. Please switch this boundary to an explicit safe-attribute policy (and sanitize URLs to origin), or remove/redact all content-bearing fields before serialization, with regression tests covering `response`, transcript previews, and credential-bearing URLs.

IanLindsley (commented):
1 high-severity issue found. The OTLP attribute redaction permits private user content and credential-bearing URLs to leave the worker, so this needs correction before merge. GitHub does not allow the PR author to submit a formal REQUEST_CHANGES review, so this comment carries the blocking verdict.

src/services/telemetry/logs.ts
```diff
@@ -0,0 +1,421 @@
...
+  'mode',
+  'mode_name',
+  'status',
+  'status_code',
+  'state',
+  'phase',
+  'step',
+  'stage',
+  'reason',
+  'error',
```
IanLindsley: **High — Do not export raw `error`/`stack` values.** This leaves the original content-egress finding partially open: `createRequestLogger.error` injects every caught exception's message and stack into these fields, and several error boundaries consume untrusted upstream text (for example, `MCPError` embeds the remote JSON-RPC `data.error.message`, then `mcp_tool_call_error` logs it). An upstream can echo tool arguments, user content, or a signed URL inside a sentence; `urlToOrigin` only sanitizes a value that is itself a URL, and truncating to 512 characters does not redact it. The collector also has no `error`/`stack` scrubber, so these values reach both sinks raw. Please summarize/drop these fields for OTLP or replace them with vetted error type/code attributes; keep full diagnostics on the existing console path if needed. Add regressions for content and an embedded credential-bearing URL inside an error message.

src/services/telemetry/logs.ts
```diff
@@ -0,0 +1,421 @@
...
+  '_type',
+  '_status',
+  '_code',
+  '_format',
+  '_reason',
+  '_mode',
+  '_language',
+  '_phase',
+  '_state',
+  '_key',
```
IanLindsley: **High — `_key` values are not safe structural metadata.** Current reachable attributes such as `audio_key`, `voice_audio_key`, `inbound_voice_key`, and `voice_submission_key` contain exact R2 locations like `audio/{org}/{user_id}/{uuid}.opus` and `voice-submissions/{org}/{chat-or-user}/{speaker-or-user}/{uuid}.ogg`. Passing the suffix raw leaks identifiers and precise audio-object locations; the collector hashes only an attribute named `user_id`, so it cannot pseudonymize IDs embedded in these strings. This also contradicts the existing trace-governance assumption that audio keys never egress. Please remove `_key` from the generic suffix allow-list and summarize/hash/drop these values, with a regression using a realistic user-bearing audio key.

IanLindsley (commented):
Re-review: the default string, nested-value, and whole-URL paths are fixed, but 2 high-severity privacy gaps remain in explicit raw-value exceptions. GitHub does not allow the PR author to submit REQUEST_CHANGES, so this COMMENT carries the blocking verdict.

IanLindsley (commented):
Re-review complete: 0 critical, 0 high, 0 medium, and 0 low issues remain. All issues addressed. GitHub does not allow the PR author to submit an APPROVE review, so this clean verdict is recorded as COMMENTED.

Issue #315, opened 2026-07-20
Wire translation-helps-mcp v2 into staging BTS


elsylambert:
The v2 server lives at:

https://translation-helps-mcp-v2.unfoldingword.workers.dev/v2/

It needs to be linked to the custom domain:

http://tc-helps.mcp.servant.bible/v2/

The deployment is a separate Cloudflare Worker named translation-helps-mcp-v2 (this separate-worker structure is what the AI produced when building it).

**Scope**

-  Point tc-helps.mcp.servant.bible/v2/ at the translation-helps-mcp-v2 worker

-  Update staging BTS config to use the v2 endpoint

-  Verify staging BTS resolves translation helps through v2

PR #326, opened 2026-07-23
fix(mcp): extract all text blocks from MCP tool results

IanLindsley:
## Problem

QA (Elsy, 2026-07-22) reported staging BTS returns *"summary counts rather than full content"* for translation notes and falls back to Aquifer.

Root cause (proven by direct server probes + staging logs): the translation-helps **v2** server returns tool results as **two** text content blocks — a short summary (`23 note(s) for Psalm 23 [kv]`, 28 chars) followed by the full payload (15,216 chars of note bodies) — and both of the worker's extraction paths keep only the **first** text block:

- `extractText` — `src/services/mcp/streamable-http-client.ts`
- `extractTextContent` — `src/services/mcp/discovery.ts` (json-rpc path)

The LLM literally received `23 note(s) for Psalm 23 [kv]` and nothing else — which is word-for-word the QA screenshot. Every earlier server (v1 tc-helps, ptxprint, yaapi) packed its whole result into one block, so the bug stayed latent; the 31/31 "successful" tool calls in staging logs confirm the failure was silent data loss, not a transport error.

## Fix

Join **all** text blocks with a blank line in both extractors; keep the `JSON.stringify(content)` fallback when no text blocks exist.

Deliberately unchanged: `structuredContent` stays ignored (no consumer exists); size caps untouched (json-rpc caps raw bytes pre-extraction; streamable-http's post-extraction cap now correctly sees the full text). Single-block responses return verbatim, preserving the ptxprint macro-tool `JSON.parse` contract. Downstream is safe — the orchestrator JSON-stringifies and truncates tool results at 32KB before they reach the model.

## Testing

- New `tests/unit/mcp-multiblock-extraction.test.ts`: both transports (mock streamable-HTTP session + stubbed json-rpc fetch) × {multi-block join, single-block verbatim, no-text-block fallback}. Verified red before the fix (multi-block cases returned only the summary) and green after.
- Full suite, lint, typecheck, architecture, build, and both audit gates pass.
- Post-deploy staging E2E planned: real chat request for Psalm 23 translation notes must contain actual note content; cf-logs to confirm no Aquifer fallback.

## Notes

- Version bumped to 2.31.1 (patch).
- Companion (non-code) fix, being applied separately to staging KV: the `PROMPT_OVERRIDES` slots still reference v1 tool names (`fetch_*`) that don't exist on v2 — being updated to v2 names (`get_*`). Tracked in #315.


IanLindsley (commented):
Review complete at 14ee02a. Critical: 0; high: 0; medium: 0; low: 0. The multi-block extraction change preserves content order and single-block behavior across both transports, with focused regression coverage. No inline findings.

PR #327, opened 2026-07-23
feat(telemetry): M3 — manual withSpan spans over non-fetch seams

IanLindsley:
## What & why

Third worker OTel milestone (M3). M1 auto-spans every outbound `fetch` + the request/DO root; M2 tees logs to OTLP. M3 fills the remaining gap: the **in-worker, non-`fetch` seams** (orchestration loop, code-execution phases, DO lifecycle, binding I/O) that are otherwise a black box between the auto fetch spans. A staging chat request now produces a full span **tree** with `status=error` landing on the correct child span.

Part of #309. Next: M4 (bounded metrics), then #324 infra/prod split (gated behind M3+M4 per QA sequencing).

## Redaction (same fail-closed posture as M2, by construction)

The hard part, and where M2's Codex loop spent two privacy rounds. Approach:

- **Shared policy, not a second one.** M2's attribute classifier is lifted out of `logs.ts` into a new `telemetry/redact.ts` (`buildSafeAttributes` / `attributeValueFor`). Both logs **and** spans go through it — spans cannot leak content a log wouldn't. `buildLogAttributes` output is unchanged (behaviour-preserving lift; existing logs tests stay green).
- **Never record the raw exception.** `withSpan`/`withSpanSync` record only the bounded `error_name` + `status=error` (no message) on throw, and never call `span.recordException` (which would write `exception.message`/`exception.stacktrace` — untrusted upstream text like an MCPError's remote message).
- **Static span names** (no ids/content). URLs → origin, unknown string keys → `string(<len>)`, objects → keys+types, sensitive keys → `[REDACTED]`.
- `redactSpan` hardened to also drop `exception.*` at export (defense in depth).

## Seams wrapped

- **orchestrator.ts**: `orchestration` loop → per-`iteration` → per-`tool_dispatch` (tool errors are caught-and-returned to the model, so the span is marked via `recordSpanError`).
- **quickjs-executor.ts**: `code_exec` → `code_exec.eval` + `code_exec.pending_calls`. Also closes a genuine silent catch — `checkExecutionError` now logs `vm_error_state_unreadable` instead of silently assuming no error.
- **r2-storage.ts / memory/store.ts**: R2 put/get + memory read/write/clear — binding I/O the library does **not** auto-instrument. Only bounded sizes on attrs; keys (which embed ids) never egress.
- **user-do.ts**: `do.enqueue` (fetch context only — the alarm-drained dequeue path cannot export spans, CF 1003, and stays log-only, covered by the tail worker).

## Safety / rollout

- **No-op when telemetry disabled**: non-recording spans, nothing egresses. Prod has no OTEL secrets + deploys are manual-dispatch → prod is unaffected. Live on staging on PR deploy (like M1/M2).
- `getTracer` is fetched per call so it binds to the provider `instrument()` registers per invocation.

## Tests

18 new unit tests (`redact` fail-closed policy; `span` records `error_name`/status without leaking message/stack, re-throws, ends span, no-op-safe). **836 total pass.** Type check, architecture (no new cycles), lint (0 errors), and build all green via pre-commit.

## Verification (post-merge, staging)

Send a chat request → OpenObserve shows the full span tree; force a tool error → `status=error` + `error_name` on the right span; whole-buffer scan = 0 hits for content / user·chat·speaker ids / R2 paths / `url.path` / `url.query` / `exception.message`.


src/services/telemetry/config.ts
```diff
@@ -83,6 +89,8 @@ export function redactSpan(span: MutableSpan): void {
   delete attrs['do.id.name'];
   delete attrs['db.statement'];
   delete attrs['db.cf.kv.key'];
+  delete attrs['exception.message'];
```
IanLindsley: **[high] Redact exception event attributes, not only span attributes**

In the pinned `@microlabs/otel-cf-workers` rc.52 implementation, `SpanImpl.recordException()` calls `addEvent("exception", attributes)`, so `exception.message` and `exception.stacktrace` live under `span.events[*].attributes`, not `span.attributes`. These deletes therefore do nothing for exceptions recorded automatically by the instrumented request/DO handlers, and untrusted messages and stacks still reach OTLP. Please redact the exception events before export (preserving `exception.type` if desired) and add a regression test whose span contains an `events` entry.

IanLindsley (commented):
Review result: 1 high-severity finding, 0 medium, 0 low. The exception redaction does not reach OpenTelemetry exception events, so the privacy guarantee is not yet met. Changes are required before merge. Submitted as COMMENT because GitHub does not allow the PR author account to request changes on its own PR.

IanLindsley (commented):
Re-review result: 0 critical, 0 high, 0 medium, 0 low. The prior exception-event redaction finding is fixed in 9ead884 and covered by regression tests. All issues addressed. Submitted as COMMENT because GitHub does not allow the PR author account to approve its own PR.

src/services/telemetry/config.ts
IanLindsley: Verified in 9ead884: redactSpan now strips exception.message and exception.stacktrace from span event attributes, and the regression test covers the rc.52 recordException shape. Resolved.

PR #328, opened 2026-07-23
feat(telemetry): M4 — bounded-label OTLP metrics

IanLindsley:
## M4 — Comprehensive bounded metrics (the last worker telemetry milestone)

Closes the worker-side slice of #309. Adds a bounded-label counter/histogram set over the same seams M1/M2/M3 already instrument, plus the auth/rate-limit/queue gap points.

### Why a custom stack (not the SDK's exporter)
`@microlabs/otel-cf-workers` is traces-only, and the SDK's `PeriodicExportingMetricReader` relies on a background `setInterval` timer that does not fire reliably between requests in Workers (and could not `fetch` outside a request context). So M4 mirrors the M2 logs stack:
- **`FetchMetricExporter`** — serializes with `@opentelemetry/otlp-transformer` (`JsonMetricsSerializer`) and POSTs to `<endpoint>/v1/metrics` over otel-cf-workers' `__unwrappedFetch` (raw fetch, so export is not self-traced).
- **`PerInvocationMetricReader`** — timer-free `MetricReader` that collects + exports exactly once per invocation, drained at the request/DO boundary via `waitUntil` (skips the POST when nothing was recorded).
- **DELTA temporality** — ephemeral isolates drain per-invocation; cumulative would report meaningless per-isolate resets. The backend sums the deltas.

### Cardinality discipline (the M4 equivalent of M2/M3's redaction)
`sanitizeMetricLabels` + `ALLOWED_LABEL_KEYS` fail **closed**: labels must be a bounded enum set; any key not on the allow-list is **dropped**, so an unbounded/content-bearing dimension (`user_id`, `request_id`, `chat_id`, raw error message, URL, R2 key) can never become a time series even from a call site that passes one in. A unit test scans the whole serialized payload to prove nothing leaks.

### Metrics emitted (beside existing `logger.*` calls)
`requests_total`, `auth_rejections_total`, `rate_limits_total`, `queue_entries_total`, `claude_fetch_total` + `claude_fetch_duration_ms`, `orchestration_iterations`, `tool_calls_total` + `tool_call_duration_ms`, `mcp_response_size_exceeded_total`, `code_execution_total` + `code_execution_duration_ms`, `r2_operations_total`, `memory_operations_total`, `tts_*` / `stt_*`, `ptxprint_poll_*`, `progress_webhook_*`.

### Wiring & governance
- Init/flush wired at the two isolate boundaries (worker fetch middleware + `UserDO.fetch`), beside the existing log init/flush.
- Genuine **no-op** when telemetry is disabled (prod has no OTEL secrets). Live on staging on PR deploy, like M1/M2/M3.

### Deps
- Pins `@opentelemetry/sdk-metrics` to `2.10.0` (matches in-tree `resources`/`core`) + a `pnpm.override` so otlp-transformer's floor of `2.0.0` dedupes to a single copy. Both audit gates stay clean (`--prod` zero, `--audit-level=high`).

### Verification
- `pnpm check` / `lint` (0 new) / `architecture` (no cycles) / `format:check` / `build` all pass.
- 12 new unit tests (856 total): fail-closed sanitization, DELTA selection, no-op-when-disabled, no-empty-POST, whole-payload leak scan, non-throwing export failure.

Bumps to **2.33.0**. Next after merge: #324 infra/prod (gated on M3+M4 landing in staging).


IanLindsley (commented):
Review result: 2 medium-severity findings and 1 low-severity finding. The metrics lifecycle currently loses measurements from background/alarms, the label sanitizer does not actually bound label values, and the response-size counter omits the default JSON-RPC path.

src/durable-objects/user-do.ts
```diff
@@ -317,6 +325,7 @@ export class UserDO {
       // are covered by the tail worker on isolate death. Full DO-lifecycle log
       // coverage (SSE lifetime, alarm) lands in M3 with the withSpan seams.
       flushLogTelemetry((promise) => this.state.waitUntil(promise));
+      flushMetricTelemetry((promise) => this.state.waitUntil(promise));
```
IanLindsley: **[medium] Flush after the work that emits these metrics.** This `finally` runs as soon as `fetch()` returns, but `processImmediateCallback`, `processImmediateSSE`, and the final-request queue drain all continue in background closures after that point. Their orchestration/tool/audio/storage metrics are therefore recorded after this flush and remain only in this in-memory provider; unlike logs/spans, this custom reader has no tail-worker export path, so a quiet DO eviction loses them. `alarm()` also neither initializes nor flushes metric telemetry, which drops the same measurements for alarm-drained entries. Please add a metrics lifecycle around alarm invocations and flush again when each background processing/drain promise completes (with tests covering post-response and alarm emissions).

src/services/telemetry/metrics.ts
```diff
@@ -0,0 +1,384 @@
...
+ * - number/boolean pass through; any other type is dropped (cannot be a bounded label).
+ */
+export function sanitizeMetricLabels(labels: MetricLabels): Record<string, LabelValue> {
+  const safe: Record<string, LabelValue> = {};
+  for (const [key, value] of Object.entries(labels)) {
+    if (!ALLOWED_LABEL_KEYS.has(key)) continue;
+    if (value === undefined || value === null) continue;
+    if (typeof value === 'string') {
+      safe[key] =
+        value.length > MAX_LABEL_VALUE_LENGTH ? value.slice(0, MAX_LABEL_VALUE_LENGTH) : value;
```
IanLindsley, line 257: **[medium] Bound values, not only keys and string length.** Truncating an arbitrary string to 64 characters still permits an unbounded number of distinct values/time series. Current call sites pass tenant/upstream-controlled values such as `serverId` and `toolCall.name` (plus runtime `error.name`) under allow-listed keys, so this does not provide the PR's claimed bounded-cardinality/fail-closed guarantee and can also egress content embedded in one of those names. Please validate/map values per key to explicit enums or otherwise drop/normalize dynamic dimensions; add tests showing distinct arbitrary values under an allowed key cannot produce unbounded series.

src/services/mcp/streamable-http-client.ts
```diff
@@ -183,6 +184,7 @@ export function enforceResponseSizeLimit(
   const text = typeof extracted === 'string' ? extracted : JSON.stringify(extracted);
   const size = new TextEncoder().encode(text).byteLength;
   if (size > maxResponseSizeBytes) {
+    countMetric('mcp_response_size_exceeded_total', { server: serverId });
```
IanLindsley: **[low] Count size rejections on the default transport too.** This counter is emitted only by the streamable-HTTP post-parse guard. The default JSON-RPC path throws the same `MCPResponseTooLargeError` from both `checkContentLengthHeader` and `readResponseWithSizeLimit` without incrementing it, so the metric systematically under-reports the dominant transport. Please instrument the shared error boundary or both JSON-RPC rejection sites and cover both transports.

src/durable-objects/user-do.ts
IanLindsley: Still open after 4bb92cd. The late fetch-background flushes improve that half of the finding, but the alarm path calls flushTelemetry even though this code states outbound fetch is blocked there. MetricReader.collect consumes the DELTA before the exporter attempts fetch; after that failed export, the next flush is empty, so there is nothing to drain on a later DO fetch. The tail worker also cannot reconstruct an in-memory OTLP metric payload. Please avoid collecting into a known-failing alarm export or persist/forward these measurements through a context that can actually export them.

src/services/telemetry/metrics.ts
IanLindsley: Still open after 4bb92cd. The SDK cardinality selector caps only the active collection window, and DELTA collect replaces that storage on every flush. I reproduced two non-overlapping batches of 900 values: both exported 900 points, yielding 1,800 backend series with no overflow. The cap also still allows the raw dynamic values to egress. Please normalize/drop dynamic dimensions with per-key value policies (or use another bound that persists across collection windows), and add a repeated-flush test.

src/services/mcp/streamable-http-client.ts
IanLindsley: Resolved in 4bb92cd: both JSON-RPC size-rejection sites now increment the counter, and the transport-specific reason labels are bounded.

IanLindsley (commented):
Re-review result: 2 medium-severity findings remain open and the prior low-severity finding is resolved. The alarm flush consumes metrics into a known-failing export, and the SDK series cap resets each DELTA collection window, so it does not provide the claimed backend cardinality bound.

src/durable-objects/user-do.ts
IanLindsley: Still open after `be653f8`. Removing the alarm flush avoids guaranteed DELTA loss, but it only defers export until a future DO fetch. If the alarm is the final invocation and the isolate is evicted, the in-memory provider is lost. The tail worker is not a backstop for these metrics: it only emits `worker_death` / `long_invocation` logs and never serializes this provider (and normal successful alarms emit nothing). Please add a durable/exportable alarm path or explicitly stop recording and claiming alarm metric coverage.

src/services/telemetry/metrics.ts
IanLindsley: Resolved in `be653f8`: runtime/untrusted values are clamped at source, error/chat/format value sets are explicit, and tool-call labels are restricted to the finite tools offered to Claude. The per-window cap is now correctly documented as a backstop.

src/services/telemetry/metrics.ts
```diff
@@ -0,0 +1,490 @@
...
+ * traffic. Clamping them to `other` would destroy the metric's purpose; the per-metric
+ * series cap (`MAX_SERIES_PER_METRIC`) backstops them against a within-window blow-up.
+ *
+ * `error_name` is kept in sync with `src/utils/errors.ts` (+ common JS built-ins and the
+ * Anthropic SDK error classes). A new error class not listed here degrades to `other`
+ * rather than leaking cardinality — a safe, self-correcting default.
+ */
+const BOUNDED_LABEL_VALUES: Record<string, ReadonlySet<string>> = {
+  chat_type: new Set(['private', 'group', 'supergroup']),
+  format: new Set(['ogg', 'mp3', 'wav', 'webm', 'flac', 'm4a']),
```
IanLindsley, line 194: **[low] Normalize accepted audio MIME values before clamping.** `transcribeAudio()` accepts MIME forms via `normalizeAudioFormat` but records the original `audioFormat`; this allow-list contains only bare extensions. A valid `audio/ogg`, `audio/mpeg`, uppercase `OGG`, etc. therefore becomes `format=other`, defeating the format breakdown for real clients. Normalize at the metric callsite (or cover every accepted alias here) and add a MIME-form test.

IanLindsley (commented):
Re-review result: 1 medium finding remains open, the prior cardinality finding is resolved, and 1 new low-severity metric-label accuracy issue was found. Alarm measurements still have no guaranteed export path; accepted MIME audio formats collapse to `other`.

src/durable-objects/user-do.ts
IanLindsley: Still open after `bf57180`. Omitting `this.initTelemetry()` does not disable metrics for an alarm that follows a fetch in the same isolate: `moduleMeter` / `moduleProvider` are module-level singletons and remain live after the fetch flush. `processQueueEntry()` then calls the instrumented orchestration/audio/storage paths from `alarm()`, so they still append DELTAs to that provider; without an alarm flush, those measurements again depend on a later fetch and are lost on quiet eviction. Please actually suppress metric recording for the alarm execution context (and test fetch-init → alarm emission → no retained delta), or provide a supported durable export path.

src/services/telemetry/metrics.ts
IanLindsley: Resolved in `bf57180`: `transcribeAudio()` now canonicalizes accepted MIME and mixed-case inputs with `normalizeAudioFormat()` before recording the label, so the emitted value is one of the six bare formats.

src/services/telemetry/metrics.ts
```diff
@@ -0,0 +1,492 @@
...
+ * them — the correct model for a stateless/serverless producer.
+ *
+ * CARDINALITY is the governing constraint here, the metrics analogue of the logs/span
+ * FAIL-CLOSED redaction, enforced in THREE layers (`sanitizeMetricLabels` + the reader):
+ *   1. KEY allow-list — drops any key not in `ALLOWED_LABEL_KEYS`, so an unbounded/
+ *      content-bearing label (user_id, request_id, chat_id, a raw error message, a URL,
+ *      an R2 key) can never become a dimension. (Also strips the PII the trace/log paths
+ *      strip.)
+ *   2. VALUE allow-list — for the labels whose values come from runtime/untrusted input
+ *      (`error_name`, `chat_type`, `format`), a value outside its `BOUNDED_LABEL_VALUES`
```
IanLindsley, line 34: **[low] Preserve the documented defense-in-depth for `format`.** This still says `sanitizeMetricLabels()` clamps `format` through `BOUNDED_LABEL_VALUES`, but the new patch removes that entry and relies solely on the STT callsite. Now that the callsite canonicalizes MIME/mixed-case values before sanitization, keeping the six-value sanitizer allow-list would no longer collapse valid inputs and would retain the fail-closed backstop for future/dynamic callers. Either restore that entry or update this contract and add a guard that makes raw `format` values impossible elsewhere.

IanLindsley (commented):
Re-review result: 1 medium finding remains open, the audio-format accuracy finding is resolved, and 1 new low-severity cardinality-contract issue was found. Removing alarm initialization does not disable the isolate-level meter after a prior fetch.

src/durable-objects/user-do.ts
IanLindsley: Resolved in `25b0f1b`: `runWithMetricsSuppressed()` wraps the entire alarm call tree in an `AsyncLocalStorage` context, and both emit helpers short-circuit on that context even when the isolate-level meter is already live. The added test verifies suppression survives an `await` while an outside emit still records.

src/services/telemetry/metrics.ts
IanLindsley: Resolved in `25b0f1b`: the six canonical formats are restored to `BOUNDED_LABEL_VALUES`, preserving defense in depth after source normalization, and the sanitizer tests cover both accepted and unknown values again.

IanLindsley (commented):
Re-review result: both prior telemetry findings are resolved. The newly added InfluxDB smoke-test script introduces 1 medium-severity credential-handling issue and 1 low-severity execution issue.

tools/send_metrics.sh
```diff
@@ -0,0 +1,55 @@
...
+for ((count = 0; count < MAX_LOOPS; count++)); do
+  RAND_VAL=$((RANDOM % 101))
+  # Deduct 1ms per loop so every point has a unique, sequential timestamp.
+  POINT_TIME=$((START_TIME - (count * 1000000)))
+  echo "btservant_smoke_test,source=smoke,bucket=${BUCKET} value=${RAND_VAL}i ${POINT_TIME}" >>"$PAYLOAD"
+done
+
+echo "POST ${URL}  (${MAX_LOOPS} points)"
+HTTP_CODE=$(curl -s -o "$RESP" -w '%{http_code}' -X POST "$URL" \
+  -H "X-InfluxDB-API-Token: ${TOKEN}" \
```
IanLindsley: **[medium] Keep the write token out of process arguments and shell history.** The documented `INFLUX_TOKEN=<token> ...` form can persist a literal credential in shell history, and expanding `${TOKEN}` into this `curl -H` argument puts it in curl's argv where process listings can expose it. That contradicts the assurance on line 24 and is risky for staging/production write credentials. Require a pre-exported or silently prompted token and pass the header through a protected temporary header/config file (or another mechanism that does not place the value in argv); update the usage text accordingly.

tools/send_metrics.sh
```diff
@@ -0,0 +1,55 @@
...
+# InfluxDB ingestion smoke test — verify the token↔bucket auth (JJ's Nginx) and the
+# auto-create-on-first-write behavior work, BEFORE wiring the OTel collector's influxdb
+# exporter ("opening the sluice"). Sends N line-protocol points straight to the write API.
+#
+# Adapted from JJ's send_metrics.sh: parameterized for bucket/token, prints the HTTP
+# status so success/failure is unambiguous, and uses a labeled, easily-purged measurement
+# (`btservant_smoke_test`, tag source=smoke) instead of a generic `test_measure`.
+#
+# Usage:
+#   INFLUX_TOKEN=<token> ./tools/send_metrics.sh <bucket>
```
IanLindsley: **[low] Make the documented direct invocation executable.** This file is committed with mode `100644`, so `./tools/send_metrics.sh ...` exits with `Permission denied` (reproduced on this head). Commit the executable bit (`100755`) or document `bash tools/send_metrics.sh ...` instead.

tools/send_metrics.sh
IanLindsley: Resolved in `90a2b69`: interactive use now reads the token silently into a non-exported shell variable, and curl receives the header through a `0600` temporary config file rather than argv. I verified with a fake curl that the prompted secret appears in neither the child argv nor environment and that the config is removed on exit.

tools/send_metrics.sh
IanLindsley: Resolved in `90a2b69`: `tools/send_metrics.sh` is now committed with mode `100755`, and direct execution succeeds in local validation.

IanLindsley (commented):
All issues addressed.

PR #329, opened 2026-07-24
feat(infra): fan metrics out to InfluxDB 3 via door43 (staging hose)

IanLindsley:
## Staging metrics hose → InfluxDB 3 (door43)

Adds **sink #2 to the collector's metrics pipeline**: the contrib `influxdb` exporter pointed at door43's Nginx write proxy (`https://metrics.door43.org`). Collector-only change — the worker never redeploys; traces/logs stay OpenObserve-only.

### Why this exporter shape
- **InfluxDB 3 Core has no native OTLP receiver** (line-protocol write endpoints only), so the collector converts OTLP metrics → line protocol and POSTs to the v2-compat `/api/v2/write` API.
- door43's Nginx **validates the `bucket=` query param against the token**, so `bucket: bt-servant-staging` must be exactly the name the token is mapped to. The bucket auto-creates on first write; retention is set by infra afterwards.
- Auth uses door43's **custom `X-InfluxDB-API-Token` header** via the exporter's confighttp `headers:` — the exporter's own `token` field would send `Authorization: Token …`, which the proxy doesn't read, so it stays unset.
- `org` is required by the exporter config but ignored by Influx 3 (placeholder).

### Verified
- Both token↔bucket pairs smoke-tested against the live endpoint with `tools/send_metrics.sh` (HTTP 204 on `bt-servant-staging` and `bt-servant`).
- Our metric labels are a bounded enum set (M4, #328), which maps directly onto Influx **tags** — no cardinality risk on the Influx side.

### Deploy (after merge)
1. `fly secrets set INFLUX_TOKEN=<staging apiv3_… token>` on `bt-servant-otel-collector`
2. `fly deploy` from `infra/otel-collector/`
3. Verify: staging chat request → measurements (`requests_total`, `claude_fetch_duration_ms`, …) appear in InfluxDB 3 Explorer
4. Ask infra to set retention on `bt-servant-staging` (auto-created buckets default to infinite)

Bumps to **2.33.1**.


IanLindsley (commented):
Reviewed head c7f3da0c. No critical, high, medium, or low issues found. The pinned otelcol-contrib 0.157.0 configuration validates, and an end-to-end local probe confirmed the InfluxDB v2 write path, bucket/org/precision query, custom X-InfluxDB-API-Token header, and line-protocol metric/tag conversion.

PR #330, opened 2026-07-24
fix(telemetry): call OTLP exporters' fetch detached from `this` (metrics + logs never reached collector)

IanLindsley:
## Problem

Worker **metrics never reached InfluxDB** and **logs never reached OpenObserve via OTLP** — only traces worked. Root cause found via the staging worker's tail logs:

```
TypeError: Illegal invocation: function called with incorrect `this` reference.
  at FetchMetricExporter.export
  at PerInvocationMetricReader.onForceFlush
  at MeterProvider.forceFlush
```

Both custom exporters (`FetchMetricExporter`, M4; `FetchLogExporter`, M2) stored the raw fetch as an instance property and called it as `this.fetchFn(...)`. Invoking native `fetch` as a **method** sets the receiver to the exporter instance, which the Cloudflare Workers runtime rejects with `TypeError: Illegal invocation`. Traces are unaffected because they use otel-cf-workers' own (correctly bound) exporter.

The throw is **synchronous** — before the promise chain — so the existing `.catch` never saw it. It escaped `export()` as a raw platform error, which is why none of the `otlp_*_export_*` structured logs ever fired and the failure was invisible until we read the raw tail stream.

## Fix

- Detach the fetcher into a local (`const fetchFn = this.fetchFn`) before calling it. A bare call in strict ESM passes `this === undefined`, which `fetch` accepts.
- Wrap the invocation in `try/catch` so any future synchronous throw is caught and reported as `otlp_*_export_error` instead of escaping the exporter (closes the observability gap the cf-logs investigation flagged).

Applied identically to `metrics.ts` and `logs.ts`.

## Tests

Added regression tests to both exporters (865 total, +4):
- **Detached-`this` guard** — a capturing fetch's `mock.contexts[0]` must be `undefined`, not the exporter instance (fails on the old code).
- **Synchronous-throw guard** — a fetch that throws synchronously (like the real illegal invocation) must be caught and reported (`code 1` / flush resolves), not escape.

## Verification plan (post-merge, staging)

1. Deploy to staging, send a WhatsApp message.
2. `SHOW MEASUREMENTS` on `bt-servant-staging` → expect `requests_total`, `claude_fetch_duration_ms`, etc. to appear.
3. Confirm log records land in OpenObserve.

Version: patch bump 2.33.1 → 2.33.2.


IanLindsley (commented):
Review result: 0 critical, 0 high, 0 medium, 0 low findings.

The detached fetch invocation fixes the Workers receiver mismatch, the synchronous-throw path now reports exporter failure through the callback, and the added regression tests cover both logs and metrics. Focused tests, the full 890-test suite, TypeScript, architecture, lint, and formatting checks pass.

Formal state is COMMENT because the active reviewer account is also the PR author and cannot self-approve.

PR #331, opened 2026-07-24
chore(deps): override postcss to clear high audit advisory (GHSA-r28c-9q8g-f849)

IanLindsley:
## Why

A high-severity advisory (`GHSA-r28c-9q8g-f849` — PostCSS path traversal via source-map auto-loading, vulnerable `<=8.5.17`) was published **after PR #330's CI passed**. `pnpm audit` hits the live advisory DB, so the next run on `main` failed the full-tree gate (`Deploy Staging` was skipped as a result). This is unrelated to any code change — pure timing.

`postcss` is a **dev-only** transitive of `vite` (via `vitest`); it never ships to the Worker runtime.

## Fix

Targeted `pnpm.overrides` entry `"postcss": ">=8.5.18 <9"` (resolves to `8.5.23`), matching the existing override pattern (`undici`, `esbuild`, `qs`, etc.). No gate weakening.

Verified locally:
- `pnpm audit --prod` → No known vulnerabilities found
- `pnpm audit --audit-level=high` → No known vulnerabilities found

Version: patch bump 2.33.2 → 2.33.3.


IanLindsley (commented):
Reviewed at `1088bb5`: 0 critical, 0 high, 0 medium, and 0 low findings. The override resolves every PostCSS path to 8.5.23, the frozen lockfile install succeeds, both audit gates pass, and the full local validation suite is green. All issues addressed.

PR #337, opened 2026-07-28
feat(telemetry): salted user_hash on the OTLP path (A1)

IanLindsley:
First slice of [docs/plans/production-otel.md](docs/plans/production-otel.md) — **A1**, which gates production telemetry (B3).

## Summary

- The worker egresses raw `user_id` to OpenObserve today, where the collector applies an **unsalted** digest. Telegram/WhatsApp ids sit in an enumerable space, so that is a lookup table away from reversal — and the live sink reports `rbac_enabled: false` (OSS build), so every OO login reads it with no per-stream scoping available.
- Adds `user_hash = HMAC-SHA-256(salt, "clientId:userId")` on the **OTLP branch only**, computed once per request and carried in `AsyncLocalStorage` (`crypto.subtle.sign` is async; `buildLogAttributes` and `redactSpan` are sync and on the hot path).
- **Logs** get `user_hash` *added* alongside `user_id`; **spans** get it *substituted*. Spans reach OpenObserve only, so a raw id there has no consumer. Logs keep `user_id` because `bt-servant-telemetry`'s OTLP receive route (Phase C1) needs it to compute its own hash — the collector deletes the wrong one per destination (B1/C2).
- Scope is established in the worker **and** independently inside the Durable Object. The DO is a separate isolate, so async context cannot cross `stub.fetch()`; without the DO wiring this labelled ~2 records per request and left the tool calls, orchestration, and timings blank.
- **Fails closed** everywhere: no salt, a missing identifier, or a hashing failure ⇒ no `user_hash` and never a fallback to the raw value.

## The console path is deliberately untouched

`logger.log()` serializes one `LogEntry` to `console.log` and *then* hands the same object to the OTLP sink. `bt-servant-telemetry` consumes the console side via a Tail Worker and derives its own hash from `obj.user_id`, so any drift there silently corrupts the cohort tables behind telemetry.btservant.ai — a rename nulls every hash, a reuse double-hashes and resets every cohort to zero.

`tests/unit/telemetry/console-path-invariance.test.ts` pins that contract: golden console bytes, byte-identical output with and without a sink registered (using a deliberately hostile sink that tries to rewrite the identifier in place), a frozen-entry no-mutation check, and a vendored stand-in for that repo's whitelist parser asserting all 14 `CleanEvent` fields plus the HMAC input string survive.

That harness was verified by making the exact mistake it guards against and confirming it goes red. **The first version did not catch it** — every case routed `user_id` through the data payload while the vulnerable line only fires for the two-arg `createRequestLogger` form used at `handleDORequest`. Coverage added, break re-simulated, went red as it should.

## Two salts, deliberately unrelated

`TELEMETRY_USER_ID_SALT` has no relationship to `bt-servant-telemetry`'s `PII_HASH_SALT` and needs none. That app keeps *using* its own salt — which cannot be read back out of Cloudflare, and does not need to be. Its hash lands in D1 and feeds the cohort KPIs; this one lands in OpenObserve as a correlation key. No table, query, or count contains both, so they cannot double-count. See "Why two hashes cannot double-count" in the plan.

Consequence worth knowing: OpenObserve records cannot be joined to D1 user rows by identity. Correlate on `request_id`, which is present in both.

## Deliberately not covered

`handleDORequest` (the admin `/do/...` route) receives a `userId` but no `client_id`. Hashing without the channel namespace would produce a **different** pseudonym for the same person — one human showing as two in the sink. It stays blank; wrong identity is worse than no identity. Confirmed as settled, not a follow-up.

## Test plan

- [x] `pnpm test` — 891 unit tests pass (34 new across two files)
- [x] `pnpm lint` — 0 errors across `src` and `tests`, no `eslint-disable` directives added
- [x] `pnpm check` — tsc clean
- [x] `pnpm architecture` — no dependency cycles
- [x] Harness verified by deliberately breaking the console contract and confirming red
- [x] HMAC asserted against an independently computed vector, not self-referential
- [x] Worker-side and DO-side scopes asserted to produce identical pseudonyms
- [ ] Staging bake — confirm `bt-servant-telemetry-dev` row counts and `user_hash` values are unchanged on live traffic

## Deploy note

`TELEMETRY_USER_ID_SALT` is already set on both workers. Unset ⇒ no `user_hash` egresses and everything else behaves exactly as before, so rollback is removing the secret.


src/services/telemetry/config.ts
```diff
@@ -111,6 +112,14 @@ export function redactSpan(span: MutableSpan): void {
   delete attrs['do.id.name'];
   delete attrs['db.statement'];
   delete attrs['db.cf.kv.key'];
+  // Spans reach OpenObserve ONLY — unlike logs, nothing downstream needs the raw id — so
+  // here we SUBSTITUTE rather than emit both. Outside a pseudonym scope the id is simply
+  // dropped (fail closed); a span without a correlation key beats a leaked identifier.
+  if (attrs['user_id'] !== undefined) {
+    delete attrs['user_id'];
+    const pseudonym = getUserPseudonym();
```
IanLindsley: **[medium] Persist the pseudonym before deferred span export.** `RedactingSpanExporter.export()` runs after spans end, not when the span is created. On the worker path, the handler leaves `withUserPseudonym` before the instrumented root span ends, so `getUserPseudonym()` is undefined here. In addition, no current span producer sets a literal `user_id`; auto-instrumented DO identity is carried in the `do.id*` fields deleted above. Consequently current spans receive no `user_hash`, despite this PR's span-correlation contract. Please attach/store the pseudonym on spans (or by trace) while the request scope is active and add a regression test that exercises the real deferred exporter lifecycle.

src/services/telemetry/pseudonym.ts
```diff
@@ -0,0 +1,107 @@
...
+ */
+export async function withUserPseudonym<T>(
+  env: Env,
+  clientId: string | undefined,
+  userId: string | undefined,
+  fn: () => Promise<T>,
+  onError?: (error: unknown) => void
+): Promise<T> {
+  const salt = env.TELEMETRY_USER_ID_SALT;
+  if (!salt || !clientId || !userId) return fn();
```
IanLindsley: **[medium] Clear an inherited pseudonym on fail-closed paths.** Calling `fn()` directly does not establish "NO pseudonym"; `AsyncLocalStorage` inherits any outer store. There is a real nested path when a group DO drains user B's queue entry from background work created in user A's `handleUnifiedChat` scope. If B's hashing fails (the catch branch has the same problem), B's processing runs with A's `user_hash`, creating cross-user attribution instead of failing closed. Explicitly run both fallback branches in a cleared/undefined store and cover a nested-scope failure in tests.

src/durable-objects/user-do.ts
```diff
@@ -780,6 +811,28 @@ export class UserDO {
   // ── Queue entry processing (called by alarm or drainQueue) ────────────────────
 
   private async processQueueEntry(entry: InternalQueueEntry, logger: RequestLogger): Promise<void> {
+    // Queued work drains from alarm(), a SEPARATE invocation from the fetch() that enqueued
+    // it — so the scope established in handleUnifiedChat is long gone. Re-establish from the
+    // stored entry body; same salt, same inputs, same user_hash as the rest of the request.
+    return withUserPseudonym(
```
IanLindsley: **[medium] Scope the per-entry log to the dequeued user.** `drainQueue()` emits `drain_queue_entry` before entering this wrapper. In a group/supergroup DO, queued entries can belong to different users; background draining retains the initiating user's async scope. Therefore a dequeued user B entry is logged with user A's `user_hash` while carrying B's `message_id`. Move that log inside the entry's pseudonym scope (or clear/re-establish the scope before emitting it) and add group-queue coverage.

src/index.ts
```diff
@@ -1757,44 +1758,80 @@ async function handleChatRequest(
...
+      (error: unknown) =>
+        logger.warn('user_pseudonym_failed', {
+          error: error instanceof Error ? error.message : String(error),
+          transport,
+          client_id: body.client_id,
+        })
     );
-    const response = await timePhase(timing, 'do_fetch', () => stub.fetch(doRequest));
+  } catch (error) {
+    return respondToChatRequestError(error, transport, timing, logger);
```
IanLindsley, line 1764: **[medium] Keep request-error logging inside the pseudonym scope.** When `dispatchChatRequest` throws, the `withUserPseudonym` promise rejects and this `await` resumes after `AsyncLocalStorage.run` has restored the outer context. `respondToChatRequestError()` then emits `request_error` without `user_hash`. That drops correlation from the failure records where it is most valuable and contradicts the stated whole-call-tree behavior. Catch/respond inside the scoped callback (or carry the pseudonym explicitly) and test a throwing dispatch path.

IanLindsley (commented):
PR review: 0 critical, 0 high, 4 medium, 0 low. Changes are required before this is clean. GitHub does not permit the PR author to submit REQUEST_CHANGES on their own PR, so this is recorded as COMMENT; see the durable summary for the blocking verdict.

src/services/telemetry/pseudonym.ts
```diff
@@ -0,0 +1,127 @@
...
+  if (!salt || !clientId || !userId) return withoutPseudonym(fn);
+
+  let pseudonym: string;
+  try {
+    pseudonym = await hashUserId(salt, clientId, userId);
+  } catch (error) {
+    // Degrade to no pseudonym — but never silently, and never to the ENCLOSING scope's
+    // pseudonym. The caller logs it with request context; telemetry losing a correlation
+    // key must not take the request with it.
+    onError?.(error);
```
IanLindsley: **[medium] Clear the inherited scope before emitting the failure diagnostic.** The handler now runs through `withoutPseudonym`, but `onError` is invoked first, while this catch is still executing under the enclosing user's store. In the real group-drain case, a failed hash for user B therefore emits `user_pseudonym_failed` with B's `message_id`/`client_id` but user A's `user_hash`. That is the same cross-user attribution this fix is meant to eliminate. Run both `onError` and `fn` inside the cleared store, and extend the nested failure test to assert that the callback itself observes no pseudonym.

IanLindsley (commented):
Re-review of 9facb6f: 3 prior medium findings are fixed; 1 remains partially addressed. Current counts: 0 critical, 0 high, 1 medium, 0 low. Changes are still required. GitHub does not permit the PR author to submit REQUEST_CHANGES on their own PR, so this is recorded as COMMENT.

src/services/telemetry/config.ts
IanLindsley: Verified on `9facb6f`: the pseudonym is attached to the active root and manual spans while they are live, and export-time redaction now only removes raw `user_id`. This finding is addressed.

src/durable-objects/user-do.ts
IanLindsley: Verified on `9facb6f`: `drain_queue_entry` now executes inside the dequeued entry’s pseudonym scope. This finding is addressed.

src/index.ts
IanLindsley: Verified on `9facb6f`: JSON parsing is separated, and dispatch failures are caught and logged inside `withUserPseudonym`. This finding is addressed.

src/services/telemetry/pseudonym.ts
IanLindsley: Partially addressed on `9facb6f`: the handler now runs with the store cleared, but `onError` still executes before that clearing and can inherit the outer user’s hash. I left this thread unresolved and added a current-line finding for the residual.

IanLindsley: Fully addressed on `f6c1692`: both the hashing-failure diagnostic and the handler now run inside `withoutPseudonym`, and the regression test directly verifies that `onError` observes no inherited pseudonym.

src/services/telemetry/pseudonym.ts
IanLindsley: Verified on `f6c1692`: `onError` has been moved into the cleared async-local scope, and the new nested-scope test covers the diagnostic callback itself. This finding is addressed.

IanLindsley (commented):
Re-review of f6c1692: the remaining medium finding is fixed. Current counts: 0 critical, 0 high, 0 medium, 0 low. All issues addressed. GitHub does not permit the PR author to approve their own PR, so this clean re-review is recorded as COMMENT.

PR #339, opened 2026-07-29
feat(telemetry): B1 — prod ENVIRONMENT, collector parameterization, collector CI/CD

IanLindsley:
Phase **B1** of `docs/plans/production-otel.md`. Addresses **item 1** of #324 (collector CI/CD); item 2 (staging/prod split) is B2/B3, so this does not close it. Follows A1 (#337).

## Summary

- **`wrangler.toml` top-level `[vars]`: `ENVIRONMENT` `"development"` → `"production"`.** The Deploy workflow runs `wrangler deploy` with no `--env` flag, so the top-level block *is* the production config — prod telemetry would otherwise arrive tagged `service.namespace=development`. Only consumers are the three `service.namespace` sites, so nothing outside telemetry changes.
- **`attributes/redact`: `user_id` `hash` → `delete`.** This is the change that finishes the privacy job A1 started. The collector's `hash` action is an **unsalted** digest — reversible by brute force for enumerable ids (phone numbers) — and it has been shipping alongside the properly salted `user_hash` since A1. The salted pseudonym already provides joinability downstream, so the raw id is pure liability here.
- **`influxdb` exporter `bucket`: hardcoded `bt-servant-staging` → `${env:INFLUX_BUCKET}`,** so one config file serves both collector apps.
- **New `deploy-collector.yml`:** `otelcol validate` against the exact image tag pinned in the `Dockerfile` on every PR touching the collector, then `flyctl deploy --remote-only` on `main`. `debug` exporter removed from all three pipelines (kept defined but unwired for future bring-ups — `verbosity: detailed` is too expensive under real traffic).

## The non-obvious bit

`otelcol validate` does **not** protect the bucket parameterization. An unset `${env:...}` resolves to an empty string and validate exits 0 — so a collector missing `INFLUX_BUCKET` boots healthy and writes every metric to `bucket=""`. Silent loss, green deploy.

Two guards close that gap:
- `check-collector-secrets.sh` pre-flights all five required secret *names* on the live fly app before any deploy.
- Grep guards in the validate job assert the bucket stays parameterized and `user_id` stays `delete`.

A matching assertion in `deploy.yml` fails the prod deploy if top-level `ENVIRONMENT` ever drifts off `"production"` — the original mislabel was invisible for a long time precisely because prod telemetry is off.

## ⚠️ Required before merge

Merging triggers the first automated collector deploy, so both must be in place first:

1. **`FLY_API_TOKEN` repo secret** — the deploy job fails without it.
2. **`INFLUX_BUCKET=bt-servant-staging` on the `bt-servant-otel-collector` fly app** — the config now reads the bucket from this secret and the old hardcoded value is gone.

The production collector job stays **skipped** until the repo variable `FLY_PROD_COLLECTOR_APP` names a provisioned app (B3 step 1).

## Test plan

- [x] `otelcol-contrib` **0.157.0** (the exact pinned tag, downloaded locally) `validate` passes on the edited config
- [x] Same validator **rejects** a deliberately broken variant (typo'd exporter ref) — confirms the gate isn't vacuous
- [x] Confirmed the empty-`INFLUX_BUCKET` blind spot exists (validate exits 0), which is why the secrets pre-flight was added
- [x] `check-collector-secrets.sh` exercised against a stubbed `flyctl`: all-present passes; missing `INFLUX_BUCKET` fails; multi-missing lists every one
- [x] Each grep guard run against its own regression (bucket re-hardcoded → fails; `user_id` reverted to `hash` → fails; current config → passes). First version had a lookahead-backtracking bug that matched everything; fixed and re-verified.
- [x] `ENVIRONMENT` assertion passes on current `wrangler.toml`, fails on a reverted copy
- [x] `wrangler build` with no `--env` reports `env.ENVIRONMENT ("production")` — wrangler confirming the premise
- [x] `pnpm test` — 923 passed / 51 files
- [x] `pnpm lint` (0 errors), `pnpm check`, `pnpm format:check`, `pnpm architecture`, `pnpm build`
- [ ] CI green on this PR — the new `validate` job runs here for real

## Not in scope

`infra/openobserve/**` gets no CI/CD; the sink stays a manual `fly deploy`. Per plan, B1 covers the collector only.


.github/workflows/deploy-collector.yml
```diff
@@ -0,0 +1,144 @@
...
+      # and exits 0, so a missing INFLUX_BUCKET would ship a collector that starts fine
+      # and writes metrics to bucket="" forever. Guard the config side here; the deploy
+      # jobs guard the fly-secret side.
+      - name: Assert no unparameterized env-specific values
+        run: |
+          set -euo pipefail
+          cfg=infra/otel-collector/otel-collector-config.yaml
+          # NB: the lookahead must sit directly after the colon. Putting a \s* BEFORE it
+          # lets the regex backtrack to zero-width and match anything.
+          if grep -nP '^\s*bucket:(?!\s*\$\{env:)' "$cfg"; then
```
IanLindsley: **[medium] Assert the required values positively and exactly.** These negative greps pass when the expected setting is absent or merely starts with `${env:`. I reproduced green results for `bucket: ${env:INFLUX_BUKCET}`, for a config with no `user_id` action, and for a multiline `user_id` / `action: hash` rule. The first case recreates the silent empty-bucket loss this gate is meant to prevent; the latter two can send the raw identifier downstream. Parse the YAML or use exact positive assertions that fail unless the bucket is `${env:INFLUX_BUCKET}` and exactly one `user_id: delete` action exists.

.github/workflows/deploy-collector.yml
```diff
@@ -0,0 +1,144 @@
...
+          fi
+          if grep -nP '^\s*-\s*\{\s*key:\s*user_id,\s*action:(?!\s*delete\b)' "$cfg"; then
+            echo "::error file=$cfg::user_id must be deleted, not hashed — the collector's hash is unsalted"
+            exit 1
+          fi
+
+  deploy-staging:
+    name: Deploy Collector (Staging)
+    needs: [validate]
+    if: github.event_name != 'pull_request'
```
IanLindsley: **[medium] Keep deployment restricted to `main`.** `workflow_dispatch` lets a caller select another branch, and this predicate treats every non-PR event as deployable, so an unmerged branch can ship to staging (and later production once `FLY_PROD_COLLECTOR_APP` is set). This also defeats the concurrency guarantee because the group includes `github.ref`, allowing two dispatched branches to deploy the same app concurrently. Require `github.ref == 'refs/heads/main'` in both deploy-job conditions and use a shared/per-app concurrency key.

.github/workflows/deploy-collector.yml
```diff
@@ -0,0 +1,144 @@
...
+    runs-on: ubuntu-latest
+    environment: staging
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: superfly/flyctl-actions/setup-flyctl@master
```
IanLindsley: **[medium] Pin the deployment action to an immutable commit.** `@master` can change without a repository commit or review. This action installs the `flyctl` binary that later receives `FLY_API_TOKEN`, so a compromised/mistaken upstream update can capture the deployment credential; this repository also currently gives workflow tokens write permission by default. Pin both occurrences to a reviewed full SHA and set the workflow/job `permissions` to the minimum needed (at least `contents: read`).

IanLindsley (commented):
Request changes: three medium-severity findings remain. The regression assertions do not prove the exact privacy/bucket invariants, manual dispatches can deploy non-main refs (and bypass the intended concurrency lock), and the Fly setup action is mutable while preparing jobs that later expose deployment credentials. The required `FLY_API_TOKEN` and staging `INFLUX_BUCKET` are also not configured yet. This review is submitted as COMMENT only because GitHub does not allow the PR author account to set REQUEST_CHANGES on its own PR.

.github/workflows/deploy-collector.yml
```diff
@@ -0,0 +1,156 @@
...
+    concurrency:
+      group: deploy-collector-app-staging
+      cancel-in-progress: false
+    steps:
+      - uses: actions/checkout@v4
+
+      # Pinned to a reviewed commit, not @master: this action installs the flyctl
+      # binary that is handed FLY_API_TOKEN two steps later, so a mutable upstream
+      # tag is a path to the deployment credential. Bump deliberately.
+      - uses: superfly/flyctl-actions/setup-flyctl@ed8efb33836e8b2096c7fd3ba1c8afe303ebbff1 # master @ 2026-04-08
```
IanLindsley, line 117: **[medium] Pin the installed `flyctl` version too.** The action source is now immutable, but this pinned action's `action.yml` declares its `version` input default as `latest`. With no `with: version: ...`, every deploy still downloads an unreviewed CLI release and then hands that executable `FLY_API_TOKEN`, so the credential-handling dependency remains mutable. Set an explicit reviewed version on both uses (and bump it deliberately alongside the action SHA).

.github/workflows/deploy-collector.yml
```diff
@@ -0,0 +1,156 @@
...
+# The M0 lesson: a config the collector rejects crash-loops the fly app on boot, and
+# fly happily rolls it out anyway. So we validate with the EXACT pinned image before
+# any deploy, and we pre-flight the fly secrets the config depends on.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'infra/otel-collector/**'
+      - '.github/workflows/deploy-collector.yml'
```
IanLindsley, line 22: **[medium] Include the workflow helper scripts in both path filters.** `.github/scripts/assert-collector-invariants.py` is now the privacy/data-loss gate and `check-collector-secrets.sh` is the deploy preflight, but a PR or main push changing only either script does not run this workflow. That lets the load-bearing checks change without collector validation and defers execution of a broken preflight until an unrelated later config deploy. Add both script paths under `push.paths` and `pull_request.paths`.

IanLindsley (commented):
Request changes: the exact invariant assertions and main-only/per-app concurrency fixes are verified, and the action source plus workflow permissions are now pinned/restricted. Two medium issues remain: the pinned setup action still installs `flyctl` from its mutable `latest` default, and the path filters omit both load-bearing helper scripts. Submitted as COMMENT because GitHub does not allow the PR author account to set REQUEST_CHANGES on its own PR.

.github/workflows/deploy-collector.yml
IanLindsley: Fixed in 8b4b104. The parsed positive assertions reject all three original bypasses; I also verified missing/hashed user_id, an unwired redaction processor, and a wired debug exporter locally.

.github/workflows/deploy-collector.yml
IanLindsley: Fixed in 8b4b104. Both deploy jobs now require refs/heads/main, and the per-app concurrency groups serialize runs targeting the same app.

.github/workflows/deploy-collector.yml
IanLindsley: Partially fixed in 8b4b104: the action source is pinned and permissions are reduced, but the pinned action manifest defaults its version input to latest, so the flyctl executable that receives FLY_API_TOKEN remains mutable. I left this thread open and anchored the residual issue on the current line.

IanLindsley: Fully fixed in 7fb44ad. The action source, workflow permissions, and installed flyctl version are all pinned/restricted; v0.4.75 exists as a non-prerelease upstream release and matches the locally verified build.

.github/workflows/deploy-collector.yml
IanLindsley: Fixed in 7fb44ad. Both action uses now set version: 0.4.75, so the executable receiving FLY_API_TOKEN is reviewed and immutable until an intentional bump.

.github/workflows/deploy-collector.yml
IanLindsley: Fixed in 7fb44ad. Both helper-script paths are present under push.paths and pull_request.paths.

IanLindsley (commented):
All issues addressed. The installed flyctl version is pinned in both deploy jobs, both helper scripts trigger the workflow for push and pull_request events, all prior findings are resolved, operational prerequisites pass, and the current head is clean and mergeable. Submitted as COMMENT because GitHub does not allow the PR author account to approve its own PR.

PR #342, opened 2026-07-29
docs(plan): record B1 shipped; stamp B2/B3 carry-forward + issue links

IanLindsley:
Docs only. Captures what B1 (#339) established so B2/B3 can't lose it, and links the plan to the new tracking issues.

Companion to the two issues filed alongside this: **#340** (B2) and **#341** (B3), which carry the same details as actionable checklists.

## Summary

- Marks **B1 ✅ SHIPPED** (`1ed38ef`, v2.35.2) in the plan of record, with the deploy-verification result.
- Adds a **carry-forward block** for the three facts B2/B3 depend on:
  1. The deploy jobs' `environment: staging` / `environment: production` declarations are **load-bearing**. GitHub resolves environment-scoped secrets over repo-level ones, so the prod job picks up a *different* `FLY_API_TOKEN` under the same name with **no workflow edit** — that's how two narrowly-scoped tokens serve one workflow. Removing them forces either a second secret name plus a workflow change, or one org-wide token, which was explicitly rejected.
  2. `FLY_API_TOKEN` is **app-scoped to the staging collector** and cannot deploy prod; B3 mints its own. **Expires ~2027-07-29.**
  3. Three deliberate version pins — including that pinning the `setup-flyctl` action SHA does **not** pin the `flyctl` binary it installs.
- Links **B2 → #340**, **B3 → #341**.
- Adds B3's full credential checklist (new app-scoped token, store on the `production` environment, verify it can read secrets and cannot see staging, set `FLY_PROD_COLLECTOR_APP`, matched bucket/token pair, prefer `--stage`).

## One design change proposed for B2

Move `INFLUX_BUCKET` out of Fly secrets into `[env]` in each app's `fly.toml`. It's a bucket *name*, not a secret — and as a Fly secret it's the single value distinguishing the two environments while being invisible and unreviewable (`fly secrets list` shows only a digest). In `fly.toml` it becomes a code-reviewed diff. Only the door43 *token* needs to stay secret.

## Test plan

- [x] `pnpm format:check`, `pnpm lint`, `pnpm check`, `pnpm architecture` pass
- [x] `pnpm test` — 923 passed / 51 files
- [x] `pnpm build` clean
- [x] No code paths touched — `docs/plans/production-otel.md` plus the version bump only
- [ ] CI green on this PR


docs/plans/production-otel.md
```diff
@@ -165,7 +165,30 @@ cumulatively from it. If we ever suspect a leak, rotating the worker's salt is c
...
+2. **`FLY_API_TOKEN` is app-scoped to `bt-servant-otel-collector` only** (verified: it cannot
+   see other apps in the org). **It cannot deploy the prod collector** — B3 mints its own.
+   It **expires ~2027-07-29**; collector deploys go red when it lapses.
+3. **Three deliberate version pins** — collector image (`Dockerfile ARG OTELCOL_VERSION`),
+   `setup-flyctl` action commit SHA, and the `flyctl` CLI version (the action's `version:`
+   input, whose default is `latest`). Pinning the action alone still installs an unreviewed
+   binary that receives the deploy token. See the table in `infra/otel-collector/README.md`.
+
+Also note `.github/scripts/assert-collector-invariants.py` **fails if `debug` is wired into
+any pipeline** — intentional, and relevant to B3 step 4, which temporarily wires it.
```
IanLindsley: **[medium] Document an executable path for the temporary `debug` deploy.** Every collector deploy job depends on `validate`, and `validate` runs `assert-collector-invariants.py`, which rejects `debug` in any pipeline. B3 step 4 therefore cannot be deployed through the workflow as written; the remaining option is an out-of-band/manual deploy, which undercuts the CI/CD boundary B1 just established. Calling the failure intentional does not resolve that contradiction. Please define the reviewed, auditable mechanism B3 should use (for example, a tightly guarded bring-up mode or a dedicated prod config/gate) and point step 4 at it.

docs/plans/production-otel.md
```diff
@@ -218,6 +248,21 @@ Ordered, because each step proves the previous one.
...
+**Credentials for B3** (see B1's carry-forward notes above for why):
+
+- Mint a **new app-scoped** deploy token: `flyctl tokens create deploy --app <prod-collector>`.
+  The existing `FLY_API_TOKEN` cannot deploy prod.
+- Store it as `FLY_API_TOKEN` **on the `production` GitHub environment**, not at repo level.
+- Verify it can run `flyctl secrets list` (the pre-flight depends on it) and **cannot** see the
+  staging app.
+- Set the **`FLY_PROD_COLLECTOR_APP`** repo variable — this is what un-skips the prod deploy
+  job. It stays skipped until then, by design.
+- Give the prod app all five secrets, with `INFLUX_BUCKET=bt-servant` (not the staging bucket)
```
IanLindsley: **[medium] Keep B3 conditional on the B2 bucket design.** If B2 moves `INFLUX_BUCKET` into each `fly.toml`'s `[env]`, this instruction adds the same name back as a Fly secret. Fly secrets take precedence over `[env]`, so the supposedly code-reviewed mapping is silently shadowed and can drift. Issue #341 already includes the needed `(unless B2 moved the bucket into fly.toml [env])` caveat; mirror it here and explicitly unset any old `INFLUX_BUCKET` secret when B2 makes that move.

IanLindsley (commented):
Request changes: 0 critical, 0 high, 2 medium, 0 low. The B3 debug-only bring-up has no path through the invariant-gated deploy workflow, and the B3 credential checklist can shadow B2's proposed code-reviewed bucket mapping. Submitted as COMMENT because GitHub does not allow the PR author account to request changes on its own PR.

docs/plans/production-otel.md
```diff
@@ -212,11 +251,55 @@ Ordered, because each step proves the previous one.
...
+     -H "Authorization: Bearer $OTEL_INGEST_TOKEN" \
+     -H 'Content-Type: application/json' -d '{"resourceSpans":[]}'
+   ```
+
+   Plus `fly logs` for `Everything is ready. Begin running and processing data.` on every
+   machine. This is exactly how B1's collector was verified on 2026-07-29, and it needs no
+   config change and no exception to the invariant gate. `debug` was an M0-era aid from when
+   no sink existed yet; it is redundant now that step 5 exists.
+
+5. Confirm records land in the **prod OpenObserve UI** — a real queryable sink, strictly
```
IanLindsley: **[medium] Emit a real test record before asking the operator to query OpenObserve.** Steps 4 and 5 run before step 6 enables the production worker, and both curls in step 4 send `{"resourceSpans":[]}`. The 200 response proves receiver/auth handling, but an empty batch produces no span for OpenObserve, so step 5 has nothing to find and cannot prove the collector→sink path. Add a valid non-empty synthetic OTLP span/log (ideally with a unique smoke-test marker) to step 4, or otherwise reorder the runbook so step 5 has an explicit record source without prematurely enabling live production telemetry.

IanLindsley (commented):
Re-review: 0 critical, 0 high, 1 medium, 0 low. Both prior findings are addressed, but the replacement B3 verification sequence still cannot prove collector-to-OpenObserve delivery before enabling the production worker: both auth probes submit an empty OTLP batch, leaving step 5 with no record to query. Submitted as COMMENT because GitHub does not allow the PR author account to request changes on its own PR.

docs/plans/production-otel.md
IanLindsley: Resolved in c79e4fc: B3 no longer wires debug; step 4 deploys only through the invariant-gated workflow, and the plan documents a reviewed two-PR route if stdout is ever genuinely required. A separate follow-up finding now tracks the missing non-empty sink smoke-test record.

docs/plans/production-otel.md
IanLindsley: Resolved in c79e4fc: B2 and B3 now make INFLUX_BUCKET conditional on the chosen design and explicitly require unsetting stale secrets that would shadow the fly.toml env value.

IanLindsley (commented):
Re-review: 0 critical, 0 high, 1 medium, 0 low. The new trace helper resolves the empty-batch sink-verification finding, but the collector README's adjacent Staged bring-up section still instructs operators to wire debug and redeploy even though the invariant-gated workflow rejects that configuration. Submitted as COMMENT because GitHub does not allow the PR author account to request changes on its own PR.

infra/otel-collector/README.md
```diff
@@ -95,15 +96,24 @@ worker's `OTEL_COLLECTOR_TOKEN` secret (`wrangler secret put OTEL_COLLECTOR_TOKE
...
-fly logs   # watch the `debug` exporter print received records
+# 2. Startup line on every machine.
+fly logs --app bt-servant-otel-collector   # "Everything is ready. Begin running and processing data."
+
+# 3. A REAL span, end to end — prompts for the ingest token.
+./tools/send_trace.sh bt-servant-otel-collector.fly.dev
 ```
 
-Then confirm the record lands in the OpenObserve UI.
+Then query the OpenObserve **traces** stream for the `smoke_marker` step 3 prints.
```
IanLindsley, line 146: **[medium] Update the adjacent `Staged bring-up` instructions to match this workflow-safe path.** Lines 120–129 still tell operators to trim pipelines to `[debug]`, append `debug`, and redeploy. `assert-collector-invariants.py` rejects `debug` in every pipeline and every deploy job depends on that gate, while the updated plan explicitly forbids bypassing it with a laptop deploy. As written, the new `Prove it` section is executable but the very next runbook remains impossible through CI and revives the original deadlock. Replace that sequence with the `send_trace.sh`/OpenObserve verification, or document the reviewed two-PR exception consistently there.

docs/plans/production-otel.md
IanLindsley: Resolved in ffd03e1: tools/send_trace.sh now emits one valid non-empty OTLP span with a unique smoke marker, and B3 requires finding that exact record in OpenObserve before enabling production telemetry.

infra/otel-collector/README.md
```diff
@@ -27,9 +27,22 @@ fly secrets set \
   INFLUX_TOKEN="<door43 apiv3_… token for the bucket below>" \
   INFLUX_BUCKET="bt-servant-staging"   # prod: bt-servant
 
-fly deploy --build-arg OTELCOL_VERSION=0.157.0   # pin a stable tag; see Dockerfile
 ```
 
+**Then let CI deploy it — do not run `fly deploy` from a laptop.** Config changes go
+PR → review → merge → [the workflow](#cicd) deploys, which is the whole point of the CI/CD
+gate (#324 item 1). The one legitimate manual use is **bootstrapping a brand-new app** before
```
IanLindsley: **[medium] Do not exempt ordinary bootstrap from the CI/CD boundary.** The preceding `fly launch --no-deploy` already creates the app; after that, the operator can mint its app-scoped token, set `FLY_PROD_COLLECTOR_APP`, and let `deploy-production` perform the first deployment. There is no chicken-and-egg dependency that requires a laptop `fly deploy`. Calling bootstrap legitimate therefore bypasses the review, validation, and privacy gates for a routine operation—the exact anti-pattern #324 item 1 removed—and conflicts with B3 step 4's workflow-only instruction. Remove the bootstrap exception. If a genuine break-glass emergency path remains documented, distinguish it explicitly and require the local validation/invariant checks before deploying, plus reconciliation through a PR.

IanLindsley (commented):
Re-review: 0 critical, 0 high, 1 medium, 0 low. The staged-bring-up contradiction is fully resolved across the repository, but the newly added deployment guidance incorrectly treats bootstrap as a legitimate reason to bypass the workflow even though fly launch --no-deploy creates the app and the production job can deploy it once FLY_PROD_COLLECTOR_APP is set. Submitted as COMMENT because GitHub does not allow the PR author account to request changes on its own PR.

infra/otel-collector/README.md
IanLindsley: Resolved in 3c142b0: the staged-bring-up section, collector config comment, infrastructure overview, and plan now consistently use the invariant-gated workflow and real-span sink verification instead of wiring debug.

infra/otel-collector/README.md
IanLindsley: Resolved in 6cb10ba: routine first deployment is explicitly workflow-only, and the remaining incident-only break-glass path runs both local gates before deploy and requires immediate PR reconciliation.

IanLindsley (commented):
All issues addressed. The first-deployment path is workflow-only, the incident-only break-glass path preserves both local gates and immediate reconciliation, all prior findings are resolved, targeted validation passes, and the current head is clean and mergeable. Submitted as COMMENT because GitHub does not allow the PR author account to approve its own PR.

Issue #257, opened 2026-06-17
MCP server resource enumeration (list tool)


elsylambert:
**Priority:** P1 / right-diamond
**Suggested owner:** TBD

## Problem
Producing an *exhaustive* resource list requires each MCP server to expose a list/enumeration tool. uW-owned servers can; third-party servers (FIA, etc.) only expose what their GraphQL/API allows. Aquifer MCP is reportedly ready.

## Proposed solution
A reliable, categorized listing of resources per MCP server (or a documented gap where a partner server can't provide it).

## Scope
- [ ] Audit which MCP servers expose enumeration today (start: Aquifer ready)
- [ ] Define the listing/categorization contract (consider an hourly-updated markdown manifest)
- [ ] Document requirements to share with partners building MCPs
- [ ] Track partner asks (e.g. FIA) to expose lists

_Source: BTServant sync._

PR #343, opened 2026-07-30
feat(#257): aggregated resources endpoint with per-server adapters (item 1)

IanLindsley:
## Summary

Item 1 of #257, implementing the contract locked on unfoldingWord/bt-servant-admin-portal#230 (Q4–Q6, signed off 2026-07-13):

- **`GET /api/v1/admin/orgs/:org/resources?language=xx`** — fans out to every enabled MCP server for the org, invokes its listing tool where one exists, and returns resources normalized into canonical subject categories plus a per-server `servers[]` status block (`ok` / `unsupported` / `error`).
- **`src/types/resources.ts`** — the canonical contract the portal mirrors verbatim: `ResourceItem`, `ResourcesBySubject`, `ResourceServerReport`, `AggregatedResourcesResponse`, and `ResourcePriority` (item 2's mode field, defined now so the whole contract lives in one file). Final types also posted on #257 for the portal side.
- **`src/services/mcp/resource-listing.ts`** — capability detection by presence of a known listing tool in the (already-run) discovery manifest: `list_resources_for_language` → translation-helps adapter (JSON, near-identity), `list` → aquifer adapter (structured markdown → canonical), neither → `unsupported` (FIA). One failing server never hides another's results.

## Design notes

- **Honest degradation**: discovery failure and listing-call failure surface as per-server `error` with the message; a server with no listing tool is `unsupported`, distinctly. Matches the portal-side rendering commitments on portal#230.
- **Format-drift protection**: the aquifer parser cross-checks parsed entries against the "Found N resource(s)" header and throws loudly on mismatch instead of silently dropping resources.
- **Subject normalization**: both server vocabularies map onto canonical slugs (`bible`, `study-notes`, `translation-notes`, `lexicon`, …); unmapped terms are slugified rather than dropped, and the summary log records `known_subject_misses` so new categories are visible in observability.
- **Language vocabularies differ**: the endpoint takes IETF-style codes (translation-helps vocabulary); the aquifer adapter translates to ISO 639-3 internally.
- **Ordering**: servers processed in ascending `priority` (same comparator as the chat path); within a subject, items keep server-default order.

## Verification

- Fixtures captured **verbatim from both live production servers** (2026-07-30): translation-helps `list_resources_for_language(en)` (9 resources / 7 subjects) and aquifer `list(eng)` (35 resources / 8 display types).
- 27 new unit tests: both parsers against the live fixtures, normalization maps, capability detection, and every degradation path (discovery failure, call failure, parse failure, unsupported, disabled, empty org).
- Full gate green locally: `pnpm lint` (0 errors), `pnpm check`, `pnpm architecture`, `pnpm test` (925 passing), `pnpm build`.

## Follow-ups (tracked on #257)

- Item 2: `resourcePriority` field on `PromptMode` (type already defined here) + portal write path.
- Item 3: chat-path wiring (prompt-bias always; category filter args where the server enforces them).

Closes nothing yet — #257 stays open until items 2–3 land.


src/services/mcp/resource-listing.ts
```diff
@@ -0,0 +1,437 @@
...
+  zh: 'zhs',
+  'zh-hans': 'zhs',
+  'zh-hant': 'zht',
+};
+
+export function toAquiferLanguage(language: string): string {
+  const key = language.toLowerCase();
+  // Lookup in a module-const map with a validated language code; worst case is undefined.
+  // eslint-disable-next-line security/detect-object-injection
+  return AQUIFER_LANGUAGE_MAP[key] ?? key;
```
IanLindsley: **Medium — normalize regional IETF tags before calling Aquifer.** The endpoint explicitly accepts tags such as `es-419`, but this exact-key fallback sends `es-419` to Aquifer instead of its required ISO 639-3 code (`spa`). Aquifer can therefore return an `ok` report with zero resources even though its listing advertises Spanish localizations; `en-US`, `pt-BR`, and similar tags have the same failure mode. Please fall back through the primary language subtag while preserving explicit script-specific mappings such as `zh-Hant`, and add a regression test for `es-419`.

IanLindsley (commented):
PR review: 0 critical, 0 high, 1 medium, 0 low. The medium correctness issue should be fixed before merge. GitHub does not allow the PR author account to submit REQUEST_CHANGES, so this review is recorded as COMMENT.

src/services/mcp/resource-listing.ts
IanLindsley: Fixed in d112e39 — exact full-tag mappings (zh-hans/zh-hant) are checked first, then the primary-subtag fallback (es-419 → es → spa, en-US → eng, pt-BR → por), then unchanged passthrough for unknown primaries. Regression tests cover all three regional cases plus the script-mapping precedence and unknown passthrough.

IanLindsley: Partially fixed in `d112e39`: the new primary-subtag fallback correctly handles `es-419`, `en-US`, and `pt-BR`, but it does not preserve a known script mapping when the tag also has a region. `zh-Hant-TW` is accepted by `validateResourceLanguage`; it misses the exact `zh-hant` entry and then falls back to `zh`, producing `zhs` instead of `zht`. Please try known prefixes from most-specific to least-specific before the primary fallback (and add a `zh-Hant-TW` regression test). Keeping this medium finding open.

IanLindsley (commented):
Re-review of d112e39: 0 critical, 0 high, 1 medium, 0 low. The regional-tag fix is partial: `zh-Hant-TW` is accepted by the endpoint but maps to Simplified Chinese (`zhs`) instead of Traditional Chinese (`zht`). The medium finding remains open. GitHub does not allow the PR author account to submit REQUEST_CHANGES, so this review is recorded as COMMENT.

src/services/mcp/resource-listing.ts
IanLindsley: Fixed in 5ceeb9f — replaced the two-step lookup with BCP 47-style fallback truncation: try the full tag, then progressively drop trailing subtags (zh-hant-tw → zh-hant → zh), so script-specific mappings win before the bare primary. Also added region-implied Chinese conventions (zh-TW/zh-HK/zh-MO → zht, zh-CN/zh-SG → zhs). Regression tests cover zh-Hant-TW → zht, zh-Hans-CN → zhs, the region shortcuts, and multi-subtag non-Chinese tags (es-419-x-priv → spa).

IanLindsley: Verified in `5ceeb9f`: fallback now checks progressively shorter prefixes, so `zh-Hant-TW` resolves through `zh-hant` to `zht` before reaching bare `zh`. The new regression tests cover this case, script precedence, region-implied Chinese mappings, and multi-subtag non-Chinese fallback. This finding is resolved.

IanLindsley (commented):
Re-review of 5ceeb9f: 0 critical, 0 high, 0 medium, 0 low. The prior language-normalization finding is fully fixed and no new issues were found. All issues addressed. GitHub does not allow the PR author account to approve its own PR, so this clean review is recorded as COMMENT.

PR #344, opened 2026-07-30
feat(#340): prod OpenObserve + prod collector config (OTel phase B2)

IanLindsley:
Phase **B2** of [`docs/plans/production-otel.md`](docs/plans/production-otel.md). Follows B1 (#339). Unblocks B3 (#341).

## Summary

- **Prod configs.** `infra/otel-collector/fly.prod.toml` (`bt-servant-otel-collector-prod` → `bt-servant` bucket) and `infra/openobserve/fly.prod.toml` (`bt-servant-openobserve-prod`, `shared-cpu-2x`/4 GB, stream data in R2 via `ZO_S3_*`, volume demoted to WAL/cache). A dedicated collector config rather than B1's `--app` override, because the app now carries environment-specific `[env]` that an override cannot express. Separate OpenObserve instance because the OSS build has `rbac_enabled: false` — no per-stream scoping, so a shared sink would give every login read access to production data.
- **`INFLUX_BUCKET` moves from a fly secret into `fly.toml [env]`.** It is a bucket name, not a credential, and as a secret it was the single value separating the two environments while being invisible — `fly secrets list` prints a digest, never the value. `assert-collector-invariants.py` now parses both fly configs and asserts each app declares the bucket the reviewed `APP_BUCKETS` map assigns it. Worth a gate because a swap does not fail: door43's proxy accepts any matching token↔bucket pair, so the mistake surfaces as staging traffic quietly polluting the production metric series.
- **The invariant checker gets a self-test** (7 fixtures it must reject, wired into `validate`). It is the only thing between a PR and a collector that hashes `user_id` or crosses the buckets, and a gate that has stopped asserting still reports green.
- **Prod deploy job** resolves its app from `fly.prod.toml` and fails if `FLY_PROD_COLLECTOR_APP` disagrees; that variable degrades to a pure enable switch. The comparison reads the variable through `env:` rather than expression interpolation, so its contents cannot reach bash as code.
- **Retention.** `ZO_COMPACT_DATA_RETENTION_DAYS` is a **fallback, not a ceiling** — the compactor prefers a stream's own `stream_settings.data_retention` whenever it is > 0 and never takes the minimum, so it governs only streams with no setting of their own. All three tiers (traces 14d, logs 30d, metrics 395d) are therefore set **per-stream and read back** in B3 step 8. The env value is the longest of the three so an unconfigured new stream errs toward over-retaining cheap data rather than deleting it. **This also means staging's `= "7"` does not fix its existing 3650-day streams** — B3 step 8 covers those too. _(Corrected in `fc1900a` after review; the original framing claimed enforcement it never provided.)_
- **The OpenObserve volume is durable metadata, not cache.** `ZO_META_STORE` defaults to `sqlite` in local mode and `ZO_DATA_DB_DIR` holds the meta store plus the `file_list` index, so moving stream data to R2 does **not** make `/data` disposable — losing it orphans every byte in R2 (still billed, unfindable) along with all users and config. B3 step 1 now enables fly volume snapshots on both instances. _(Also `fc1900a`.)_
- **Docs + plan** updated for the two-environment topology, and B2.5 (CI/CD for `infra/openobserve/**`) is scheduled ahead of B3 — see below.

## Sequencing — this is PR A of two

A fly secret **takes precedence over `[env]`**, so the staging app's leftover `INFLUX_BUCKET` still shadows the new value. The pre-flight only **warns** about that today, deliberately:

1. **This PR** — merging deploys `[env]` into the machine config. The secret still shadows it with an identical value, so nothing changes behaviourally.
2. **Manual** — `flyctl secrets unset INFLUX_BUCKET --app bt-servant-otel-collector`. The restart re-reads a machine config that now has `[env]`, so the bucket stays correct.
3. **PR B** — promote the warning to a hard failure (`TODO(#340 follow-up)` in the script).

Failing in step 1 would deadlock the very deploy that installs `[env]`; unsetting first would leave the running collector writing to `bucket=""`.

## Also flagged: B2.5

`infra/openobserve/**` has no workflow, so both instances are hand-deployed. Tolerable while OpenObserve was one staging box — not once B3 creates the production sink, since B3 would have to bootstrap it with a laptop `fly deploy`, the anti-pattern #324 item 1 removed for the collector. Recorded in the plan as **B2.5, before B3**. No code for it in this PR.

## Test plan

- [x] `assert-collector-invariants.py` self-test — 7/7, including the original `user_id: hash` privacy regression
- [x] **Sabotage check**: pointed the real staging `fly.toml` at the prod bucket and confirmed the gate goes red with the annotation on the right file, then restored it. Green-on-arrival proves nothing.
- [x] Verified `::error file=` paths are repo-relative from both the CI cwd and the break-glass runbook's cwd — absolute paths silently fail to attach to a file in the PR view
- [x] All four fly configs parse (`tomllib`); workflow parses and both deploy jobs resolve the intended app via the workflow's `sed`
- [x] `pnpm check` (tsc), `pnpm architecture` (no cycles), `pnpm test` (952 passed), `pnpm build`
- [x] Version bumped 2.36.1 → 2.36.2
- [ ] `otelcol validate` against the pinned image — needs Docker, runs in CI (collector config itself is unchanged in this PR)

Refs #340, #324, #309.


infra/openobserve/fly.prod.toml
```diff
@@ -0,0 +1,70 @@
...
+  # to object storage. Without ZO_LOCAL_MODE_STORAGE the other ZO_S3_* values are read
+  # and then ignored, and everything silently keeps landing on the volume.
+  ZO_LOCAL_MODE_STORAGE = "s3"
+  ZO_S3_PROVIDER = "s3"
+  ZO_S3_BUCKET_NAME = "bt-servant-openobserve-prod"
+  ZO_S3_REGION_NAME = "auto"     # R2's region identifier
+  # ZO_S3_SERVER_URL, ZO_S3_ACCESS_KEY, ZO_S3_SECRET_KEY are fly secrets — the URL
+  # embeds the Cloudflare account id, and the keys are credentials. See README.md.
+
+  # Local volume below is now WAL + query cache only, not the retention boundary.
```
IanLindsley: **[medium] Treat this volume as durable metadata, not only WAL/cache.** In OpenObserve v0.91.0, local mode defaults `ZO_META_STORE` to SQLite, and `metadata.sqlite` plus the file-list index live under `ZO_DATA_DIR/db` even when `ZO_LOCAL_MODE_STORAGE=s3`. Losing or replacing this volume can therefore lose users/configuration and the index needed to query the R2 objects. Please correct the comments/runbooks and include an explicit backup/restore expectation for the volume (or move metadata to a durable external store).

infra/openobserve/fly.prod.toml
```diff
@@ -0,0 +1,70 @@
...
+  #
+  #     traces   14 days
+  #     logs     30 days
+  #     metrics  395 days  (bounded-label, so tiny)
+  #
+  # Set that way round on purpose. If the per-stream step is ever missed, the failure is
+  # over-retaining traces in cheap object storage — not silently deleting the metrics
+  # history, which is the one signal here that cannot be reconstructed after the fact.
+  # Applying the two overrides is a step in production-otel.md B3.
+  ZO_COMPACT_DATA_RETENTION_DAYS = "395"
```
IanLindsley, line 65: **[medium] This value is a fallback, not a ceiling.** In v0.91.0 the compactor uses `stream_settings.data_retention` whenever it is greater than zero and only falls back to `ZO_COMPACT_DATA_RETENTION_DAYS`; it does not take the minimum. An existing/future 3650-day stream override therefore bypasses 395 here (and likewise bypasses the new 7-day staging value). Please make the runbook explicitly set and verify all three stream values—including metrics=395 and the existing staging streams=7—and remove the `ceiling` claims so the intended retention is actually enforced.

IanLindsley (commented):
Review found 2 medium-severity operational correctness issues: the configured retention value does not enforce the documented ceiling, and the production volume still contains durable SQLite metadata despite being described as WAL/cache only. Changes are needed before merge. This is submitted as COMMENT because GitHub does not allow the PR author to request changes on their own PR.

infra/openobserve/fly.prod.toml
IanLindsley: Confirmed and fixed in `fc1900a` — you're right, and the comment you flagged was actively misleading.

Verified against the v0.91 docs: `ZO_META_STORE` defaults to `sqlite` in local mode, `ZO_DATA_DB_DIR` sits under `ZO_DATA_DIR`, and the `file_list` index — the thing that maps a query to the parquet objects — lives in that same meta store. `ZO_LOCAL_MODE_STORAGE=s3` moves the stream data and nothing else.

So the failure mode isn't a cold cache, it's worse than "lose users/config": **the R2 bucket gets orphaned.** The objects survive, stay billed, and become unfindable, because the index that knew they existed was the thing on the volume. That's a silent, total loss of the production history dressed up as a healthy instance.

What changed:

- `ZO_DATA_DIR` comment rewritten to say what actually lives there and what losing it costs.
- The `[[mounts]]` comment no longer reads as a sizing note — it states this needs snapshots and a tested restore.
- The `[[vm]]` comment no longer describes the disk as "WAL and cache".
- New **"The volume"** section in `infra/openobserve/README.md` spelling out both failure modes, and naming the external-Postgres meta store (`ZO_META_STORE=postgres` + `ZO_META_POSTGRES_DSN`) as the route if we ever want the instance itself to be disposable. We're not doing that yet — flagging it as a real option rather than pretending the volume is stateless.
- **B3 step 1 now enables fly volume snapshots on both OpenObserve volumes**, so the backup expectation is a checklist item in the runbook that provisions them, not a sentence in a README someone reads afterwards.
- The `fly volumes create` line in the deploy block points at the section, since that's where someone is actually acting.

infra/openobserve/fly.prod.toml
IanLindsley: Confirmed and fixed in `fc1900a`. This one was worse than a wording problem — it made part of the PR a no-op, and I'd missed that.

You're right that the compactor prefers `stream_settings.data_retention` whenever it's > 0 and only falls back to the env var, without taking the minimum. Following that through:

**The staging change did nothing.** `infra/openobserve/fly.toml` set `ZO_COMPACT_DATA_RETENTION_DAYS = "7"` to fix the instance reporting 3650. But those streams already carry 3650 as a stream setting, so they'd have gone on ignoring it indefinitely. The PR would have claimed to bound staging retention while changing nothing at all — and nobody would have noticed, because there's no signal that distinguishes "bounded" from "still ten years" short of reading the stream back.

So "remove the ceiling claims" alone wasn't enough; the runbook had to grow the step that actually enforces it:

- Every "ceiling" claim is gone from `fly.prod.toml`, `fly.toml`, `infra/openobserve/README.md`, and `docs/plans/production-otel.md`. Each now says fallback, states that the compactor never takes the minimum, and says explicitly that the env var only governs streams with no setting of their own.
- **B3 gains step 8: set and verify all three streams** — prod `traces`=14d, `logs`=30d, `metrics`=395d, and **staging's existing three at 7d**, which is the remediation for the no-op above.
- It's step **8**, last, because prod's streams don't exist until first ingest creates them — putting it in step 1 would have meant configuring streams that aren't there yet. It's appended rather than inserted because steps 4–6 are cross-referenced by number both here and in #341.
- Added a read-back to the README, since "I set it in the UI" isn't evidence:

  ```bash
  curl -s -u "$O2_USER:$O2_PASS" "https://<instance>.fly.dev/api/default/streams" \
    | jq -r '.list[] | "\(.name)\t\(.stream_type)\tretention=\(.settings.data_retention // "unset -> env fallback")"'
  ```

  Anything still showing 3650 means it didn't take.

I kept the env value at the longest tier (395), but the justification is now the accurate one: it's the default for a stream nobody has configured yet, so a new stream errs toward over-retaining cheap data instead of deleting it. That was always the real argument — I'd overstated it into enforcement it never provided.

IanLindsley (commented):
Re-reviewed the remediation commit and full PR. Both prior medium findings are resolved, no new issues were found, and focused validation plus the current CI suite pass. All issues addressed. This is submitted as COMMENT because GitHub does not allow the PR author to approve their own PR.

infra/openobserve/fly.prod.toml
IanLindsley: Verified in fc1900a: the config and runbooks now identify the SQLite meta store and file-list index as durable volume state, and provisioning explicitly requires snapshots. This finding is resolved.

infra/openobserve/fly.prod.toml
IanLindsley: Verified in fc1900a: the fallback semantics are documented accurately, B3 explicitly sets all production and staging stream retentions, and the read-back command reports each stream setting. This finding is resolved.

PR #347, opened 2026-08-03
fix(#340): fail the deploy on a shadowing INFLUX_BUCKET secret (B2 PR B)

IanLindsley:
The second and final PR of the two-part sequence #340 was split into. Follows PR #344 (`4cbdac9`, v2.36.2).

## Summary

- **The `SHADOW` warning is now a hard failure.** A fly secret takes precedence over `fly.toml [env]`, so a leftover `INFLUX_BUCKET` doesn't misconfigure the app in any *visible* way — it makes the reviewed value in git dead code while `fly secrets list` shows only a digest. Setting that secret isn't a way to override the bucket, it's a way to hide it. Now it can't ship.
- **Both problem lists are reported before exiting.** Failing on the first one found would hide the second, and an operator who fixes a missing secret shouldn't then discover a shadowed one on the next run.
- **`check-collector-secrets.sh` gets its first test.** It had none, and in PR A it grew teeth in two directions: it can now fail a deploy because a required secret is *missing* and because a banned one is *present*. A gate that quietly stops asserting waves bad configs through; one that always fails blocks every collector deploy until someone debugs CI. Neither is visible from reading the script.

## Why this is safe to tighten only now

The gate could not have been introduced in PR A — it would have blocked the very deploy that made it satisfiable:

1. **PR A** (#344) shipped `[env] INFLUX_BUCKET` and *warned* on the leftover secret. Merging deployed `[env]`; the secret still shadowed it with an identical value, so behaviour was unchanged. The live deploy printed `SHADOW INFLUX_BUCKET` exactly as designed.
2. **Manual, done 2026-08-03** — `flyctl secrets unset --app bt-servant-otel-collector INFLUX_BUCKET`. Verified after: both machines report `INFLUX_BUCKET=bt-servant-staging` sourced from `[env]`, and the collector stayed up throughout (401 on all 40 probes across a 10-minute window).
3. **This PR** — tighten the gate, now that no collector app carries the secret.

Doing it in one step would have deadlocked; clearing the secret first would have left the collector writing every metric to `bucket=""`. The plan records the general pattern, since any future "move a fly secret into `[env]`" needs the same three beats.

## One coupling this had to break

Testing the script meant decoupling it from the real fly API. It hardcoded `flyctl`, so the only way to exercise it was against a live authenticated org — which is precisely what happened the first time I ran the new test: it made a real API call against a developer session. It now calls `"${FLYCTL:-flyctl}"`, and the self-test stubs that with a fake emitting canned `secrets list --json`. No credentials, no network, so it runs in `validate` rather than the deploy jobs.

## Test plan

- [x] 7 self-test cases pass: happy path, shadowed bucket, missing required, empty list, both problems at once (asserting **both** are reported, not just the first), and a near-miss name (`OTEL_INGEST_TOKEN_OLD`) that must not satisfy a required secret
- [x] **Sabotage check**: reverted the shadow check to a warning → suite goes red on exactly that case; green again on restore. Green-on-arrival proves nothing.
- [x] Confirmed the stub is actually used (no real fly API calls) after the `FLYCTL` change
- [x] `assert-collector-invariants.py` + its self-test still green; workflow parses with all three self-test steps wired into `validate`
- [x] `pnpm check` / `architecture` / `test` / `build` via pre-commit
- [x] Version bumped 2.36.2 → 2.36.3

**On merge**, the collector redeploys and the hard gate runs against the real app for the first time. It should pass — the secret is gone. If it didn't, the deploy simply stops and the running collector is untouched.

Closes #340.


IanLindsley (commented):
Reviewed head `37bdfd7a0165909482275572c54750b3af5699c4`: 0 critical, 0 high, 0 medium, and 0 low findings. All issues addressed. Submitted as COMMENT because the active reviewer account is also the PR author, so GitHub cannot record a self-approval.

PR #348, opened 2026-08-03
feat(#345): CI/CD for infra/openobserve/ (OTel phase B2.5)

IanLindsley:
Phase **B2.5** of `docs/plans/production-otel.md`. Follows #347 (v2.36.3). Must land **before** B3 (#341) provisions the production sink.

## Summary

`infra/openobserve/**` was the last hand-deployed directory: a change to either fly toml did nothing until someone ran `fly deploy` from a laptop. Left as-is, B3 would have had to bootstrap the **production** sink that way — the anti-pattern #324 item 1 removed for the collector — and prod's deployed state would have started out unreconcilable with `main`.

- **`deploy-openobserve.yml`** mirrors `deploy-collector.yml`: validate on every PR touching `infra/openobserve/**` / the workflow / `.github/scripts/**`; deploy-on-merge for staging; prod job gated behind the `FLY_PROD_OPENOBSERVE_APP` repo variable (set in B3) and failing loudly if the variable disagrees with `fly.prod.toml`'s `app` rather than deploying to whichever won. Deploy jobs read **`FLY_OPENOBSERVE_API_TOKEN`** — the environment secrets named `FLY_API_TOKEN` are the collector apps' tokens, and fly deploy tokens are app-scoped.
- **Validate asserts what runtime never reports.** It parses `ARG OPENOBSERVE_VERSION` from the Dockerfile — deploy passes it as `--build-arg`, so the pin has one enforced source of truth and the workflow fails loudly if the ARG is ever renamed. (Honesty note vs. the issue text: the ARG carries a default, so a bare `fly deploy` was already pinned; what was missing was the loud failure and the single source of truth.) It also asserts `fly.prod.toml` keeps `ZO_LOCAL_MODE_STORAGE = "s3"`, without which every other `ZO_S3_*` value is read and **silently ignored** — stream data lands on the volume until it fills.
- **`check-openobserve-secrets.sh`** pre-flights each app before its deploy. Required secrets are per-app (staging: the two `ZO_ROOT_USER_*`; prod adds `ZO_S3_SERVER_URL` / `ZO_S3_ACCESS_KEY` / `ZO_S3_SECRET_KEY` — wrong or missing R2 credentials are not a boot failure, they surface later on a write path). The shadow gate's banned list is **parsed from the toml's `[env]` keys** rather than hand-maintained — the `INFLUX_BUCKET` lesson (#340) generalized, so adding an `[env]` key can never leave the gate behind. Zero parsed keys is itself a failure: a parser that silently returns nothing would disarm the gate while reporting green.

## Test plan

- [x] 15 self-test cases pass (stubbed flyctl, no credentials/network; runs against the **real** repo tomls so the `[env]` parser is exercised on exactly what CI feeds it): both happy paths, missing R2 credential, missing root secret, staging's set not satisfying prod, shadowed key on each app, both-problems-at-once (asserting **both** reported), near-miss name, unknown app, toml without `[env]`, missing toml path
- [x] **Sabotage check both directions**: dropped `ZO_S3_SECRET_KEY` from the prod required map → red on exactly the two cases that depend on it; disarmed the shadow gate → red on exactly the two shadow cases; green again on restore
- [x] Collector's self-tests still green (shared `.github/scripts/` triggers)
- [x] Workflow YAML parses; the new workflow's `Validate Config` job runs on this PR (its paths include the workflow file)
- [x] `pnpm check` / `architecture` / `test` (952) / `build` via pre-commit
- [x] Version bumped 2.36.3 → 2.37.0

## Before merge (manual)

The staging deploy job fires on merge and needs the staging OpenObserve deploy token to exist:

```bash
flyctl tokens create deploy --app bt-servant-openobserve
gh secret set FLY_OPENOBSERVE_API_TOKEN --repo unfoldingWord/bt-servant-worker
```

The prod token and `FLY_PROD_OPENOBSERVE_APP` wait for B3, as planned.

**Side effect on merge**: `.github/scripts/**` is also in `deploy-collector.yml`'s path globs (deliberately), so the staging collector redeploys idempotently — same as #347.

Closes #345.

## Rider: brace-expansion 5.0.8 → 5.0.9 (GHSA-rgw5-rvv9-x895)

A new high advisory landed mid-PR and turned the Security Audit gate red for every branch: 5.0.8's CVE-2026-14257 mitigation is bypassable; 5.0.9 closes it. The override is bumped and the local callable-CJS-default-export patch (still needed — 5.0.9 keeps the object-only export) is regenerated against 5.0.9. Verified: both audit gates clean, patched default export callable, `pnpm lint` exercises the ESLint→minimatch@3 require path with zero errors.


IanLindsley (commented):
Review result: 0 critical, 0 high, 0 medium, 1 low. The OpenObserve CI/CD and secret pre-flight otherwise look sound. Re-review requested after the source-map patch is corrected.

patches/brace-expansion@5.0.9.patch
```diff
@@ -1,16 +1,17 @@
...
 +++ b/dist/commonjs/index.js
-@@ -260,4 +260,11 @@ function expand_(str, max, maxLength, isTop) {
+@@ -286,4 +286,10 @@ function expand_(str, max, maxLength, isTop) {
      }
      return acc;
  }
-+// [bt-servant patch] Restore the callable CommonJS default export. Upstream 5.x
+-//# sourceMappingURL=index.js.map
+\ No newline at end of file
++//# sourceMappingURL=index.js.map// [bt-servant patch] Restore the callable CommonJS default export. Upstream 5.x
```
IanLindsley: **[low] Preserve the source-map directive when adding this patch comment.** Because upstream's file has no trailing newline, this concatenates the prose onto the `sourceMappingURL` value (`index.js.map// ...`) and makes the URL invalid. I reproduced the regression with `node --enable-source-maps`: errors now point at `dist/commonjs/index.js:89`, while unpatched 5.0.9 maps the same frame to `src/index.ts:105`. Please insert a newline before the patch comment (or put the comment before the directive) so source-mapped stack traces keep working.

IanLindsley: Fixed in 8863de8. Verified with the original `node --enable-source-maps` repro: the frame now maps to `src/index.ts:105`, while the callable default and named export both remain intact. Closing this thread.

IanLindsley (commented):
All issues addressed. Re-reviewed head 8863de8: the brace-expansion export patch now preserves the source-map directive, source-mapped errors resolve to src/index.ts, the callable CommonJS export still works, and no new issues were introduced. Formal state is COMMENT because the active reviewer account authored this PR.

PR #349, opened 2026-08-04
feat: DO identity persistence, hex-addressed admin snapshot, chat_turn telemetry

IanLindsley:
## Why

Jacky asked for the language users chat in, broken down by country. Investigation showed the runtime knows both facts on every turn but never records them anywhere queryable, and that ~540 of the 861 all-time users live in Durable Objects that cannot be attributed to anyone: the CF REST object-listing API returns one-way 64-hex ids, `idFromName` is not reversible, and the DO never stored its own name. This PR closes both gaps — it recovers the past (every DO's history becomes reachable and, going forward, attributable) and secures the future (country × language becomes a standing metric instead of a 7-day log-archaeology exercise).

## What

**Identity persistence (recover the past)**
- `UserDO` persists a `StoredIdentity` record (`do_name` mirroring the exact `idFromName` key, plus org / user_id / chat_id / thread_id / client_id / chat_type) on every chat turn, before rate limiting, so even failed turns attribute the DO. Steady-state turns skip the write (read + compare).
- New DO route `GET /identity`.
- New super-admin endpoint `GET /api/v1/admin/do/:hexId/snapshot` — opens any DO via `idFromString` and returns `{do_id, identity, history}`. Sits outside `/api/v1/admin/orgs/*`, so the global `/api/*` middleware (ENGINE_API_KEY) is the sole gate. Malformed hex → 400.

**chat_turn telemetry (secure the future)**
- One `chat_turn` log record per addressed turn: `response_language`, edge `country` (`request.cf.country`, injected by the worker as `_request_country` — the DO can't see `request.cf` across the stub boundary), `client_id`, `transport` (stamped as `_transport` so queued entries retain it), `chat_type`. Ambient short-circuited turns are deliberately excluded (no response → no response_language).
- New `chat_turns_total` counter using the already-allow-listed `language` label plus a new bounded `country` label (ISO 3166 set). `country` added to the OTLP metric label allow-list and the log redaction allow-list.

## Tests

- New e2e suite `do-identity-snapshot.test.ts`: fresh DO → null identity; chat turn persists identity (private and group DO naming); snapshot endpoint happy path via SELF; malformed hex → 400; missing auth → 401. (e2e runs in CI on Linux; Windows skips locally.)
- Unit: `language`/`country` metric labels pass `sanitizeMetricLabels`.
- `pnpm lint` / `check` / `architecture` / `test` all green (928 passed).

## Version

Minor bump → 2.37.0 (new endpoint + new telemetry surface).


src/index.ts
```diff
@@ -1781,6 +1829,10 @@ async function buildDOChatRequest(
       _org_prompt_overrides: promptOverrides,
       _org_modes: orgModes,
       _org_languages: orgLanguages,
+      // Edge country of the ORIGINAL request (gateway location for relayed
+      // clients, user location for direct web clients) — the DO cannot see
+      // request.cf across the stub boundary, so it rides in the body.
+      _request_country: (request.cf as { country?: string } | undefined)?.country,
```
IanLindsley: **Medium — this records the gateway's country, not the user's, for relayed traffic.** The PR's stated `country × language` metric therefore collapses WhatsApp/Telegram users into the gateway deployment country and cannot answer the motivating user-geography question. Please propagate a bounded, authenticated end-user country from the gateway (or name this `edge_country` and avoid presenting it as user country); otherwise the resulting metric is materially misleading.

src/index.ts
```diff
@@ -1168,6 +1174,48 @@ app.delete('/api/v1/admin/orgs/:org/users/:userId/history', async (c) => {
...
+    return c.json({ error: 'hexId must be a 64-character lowercase hex Durable Object id' }, 400);
+  }
+  try {
+    const stub = c.env.USER_DO.get(c.env.USER_DO.idFromString(hexId));
+    const limit = c.req.query('limit') ?? '100';
+    const [identityRes, historyRes] = await Promise.all([
+      stub.fetch('http://do/identity'),
+      stub.fetch(`http://do/history?limit=${encodeURIComponent(limit)}`),
+    ]);
+    const { identity } = (await identityRes.json()) as { identity: StoredIdentity | null };
```
IanLindsley: **Medium — propagate failures from both DO subrequests before decoding them.** `/identity` converts storage errors into a non-2xx JSON response, and `/history` can also return non-2xx. `fetch()` still resolves in those cases, so this code casts the error payload and returns an outer 200 with missing `identity` or an error-shaped `history`. That makes an incomplete snapshot indistinguishable from a successful one. Check both `ok`/status values and return an error (or a clearly typed partial-result status), with a test covering a failed subrequest.

package.json
```diff
@@ -84,22 +84,23 @@
       "@opentelemetry/core": "^2.8.0",
       "@opentelemetry/sdk-metrics": "2.10.0",
       "express-rate-limit": ">=8.2.2",
+      "fast-uri": ">=3.1.4 <4",
```
IanLindsley: **Low — the override still admits the vulnerable `fast-uri@3.1.4`.** The lockfile currently selects 3.1.5, so today's audit is green, but the manifest does not enforce the advisory's fixed floor despite the commit message saying it was pinned. Please change the lower bound to `>=3.1.5 <4` (and keep the lockfile aligned) so future resolutions cannot legally select 3.1.4.

IanLindsley (commented):
Review result: 0 critical, 0 high, 2 medium, 1 low. The country dimension does not represent user geography for gateway-relayed traffic, snapshot subrequest failures can be returned as successful snapshots, and the fast-uri override still permits the vulnerable release. These findings require changes and re-review; GitHub does not allow the PR author account to submit REQUEST_CHANGES on its own PR, so this review is recorded as COMMENT.

package.json
IanLindsley: Addressed in dd7d82d. The manifest now enforces `fast-uri >=3.1.5 <4`, the lockfile resolves 3.1.5, and the production audit passes.

src/index.ts
IanLindsley: Addressed in dd7d82d. `readDoSnapshot` now checks both subrequest statuses before decoding, the route returns 502 for an incomplete snapshot, and the new unit tests cover identity failure, history failure, and both failing.

src/durable-objects/user-do.ts
```diff
@@ -254,6 +253,72 @@ function logLanguagePersistenceChange(
...
+  chatType: string;
+  transport: ChatTransport | undefined;
+  userCountry: string | undefined;
+  edgeCountry: string | undefined;
+} {
+  return {
+    org: body.org ?? body.org_id ?? defaultOrg,
+    chatType: body.chat_type ?? 'private',
+    transport: body._transport,
+    userCountry: countryFromPhoneUserId(body.user_id),
```
IanLindsley: **Medium — gate phone-country inference by an identity source known to supply E.164.** This call treats every all-digit 7–15 character `user_id` as a verified phone number. The sibling Telegram gateway sends `String(from.id)` with `client_id: "telegram-gateway"`; real Telegram account IDs are numeric but are not phone numbers. Consequently, IDs such as `1234567890` and `6281319183` are currently emitted as US and Indonesia, respectively, corrupting the new user-country metric. Please restrict inference to clients whose identifier contract is E.164 (or propagate an explicit authenticated phone-derived country), and add a regression test proving a numeric Telegram ID does not produce `user_country`.

IanLindsley (commented):
Re-review result: 0 critical, 0 high, 1 medium, 0 low. The snapshot failure handling and fast-uri floor are fixed. The geography fix is incomplete because numeric non-phone identifiers, including raw Telegram account IDs, are still interpreted as E.164 phone numbers. This requires another change and re-review; GitHub does not allow the PR author account to submit REQUEST_CHANGES on its own PR, so this review is recorded as COMMENT.

IanLindsley (commented):
Re-review result: 0 critical, 0 high, 0 medium, 1 low. The E.164 client gate fully addresses the remaining telemetry correctness finding: Telegram and unknown clients fail closed, while the WhatsApp and Signal contracts are explicitly covered. One low issue remains: tests/unit/phone-country.test.ts again contains a literal NUL byte, so Git classifies the TypeScript test as binary and hides its diff. This regresses the explicit c68ce3e text-file fix; replace the byte with the six-character Unicode escape (a backslash followed by u0000) and re-review. No inline anchor is available because Git treats the file as binary.

IanLindsley (commented):
Re-review result: 0 critical, 0 high, 0 medium, 0 low. The NUL-byte regression is fixed: the final test source contains the Unicode escape, has zero literal NUL bytes, and is exposed as a normal textual GitHub diff. All prior findings remain resolved. All issues addressed.

PR #350, opened 2026-08-05
feat(infra): raise prod OpenObserve retention to 5 years (1825d)

IanLindsley:
## What

Raises production OpenObserve retention to **1825 days (5 years)** by setting the
`ZO_COMPACT_DATA_RETENTION_DAYS` fallback in `infra/openobserve/fly.prod.toml`, and records
the decision that **production runs a single retention tier via the fallback**, with no
stream carrying its own value.

## Why the three-tier plan was dropped

`docs/plans/production-otel.md` B3 step 8 called for per-stream tiers (traces 14d / logs 30d
/ metrics 395d). That did not survive contact with the running instance:

- **`metrics` is not one stream.** OpenObserve splits every OTel metric into its own stream
  and fans each histogram into `_bucket`/`_count`/`_max`/`_min`/`_sum` — ~28 metric streams
  on day one, plus a new one every time a metric is added. Per-stream retention would have to
  be re-applied forever; the `[env]` fallback covers new streams automatically.
- **Volume made the tiering moot.** 168 spans measured 0.086 MB compressed (~0.51 KB/span).
  At ~5–10k spans/day, five years is roughly **4.6–9.4 GB** — under **$0.15/month** of R2.
  Retention here is a data-policy and query-performance question, not a cost one.

## ⚠️ The trade-off this accepts

With no stream carrying its own value, the compactor evaluates this variable **live** rather
than stamping it at write time. **Lowering it retroactively deletes everything older than the
new value, across every stream, on the next compaction** — silently, with no alert and no
archive. Edits to that line are destructive and should be reviewed as such.

This is called out in all three places the value is documented.

## Also fixed (found while walking the SOE)

- `wrangler secret list --env production` **fails** — there is no `[env.production]` section;
  production is the top-level config and only `staging` is a named environment. Now documents
  `--env=""` plus the required `CLOUDFLARE_ACCOUNT_ID`, without which wrangler silently
  targets whichever account the ambient env points at.
- `/health` on `workers.dev` can serve the **previous** version from the edge for a minute or
  two after deploy, which reads as a failed deploy. Now says to cache-bust.

## Not in scope

**Staging is deliberately left at 3650 days.** Changing it was considered and declined.

## Deployment

Merging triggers `deploy-openobserve.yml`, which redeploys and **restarts** the prod
OpenObserve machine. That briefly interrupts ingest and queries; the collector's exporter
queue absorbs the gap. Increasing retention is safe — nothing is deleted, and no production
data has aged out yet (ingest began 2026-08-05).

## Verification

```bash
# After deploy — every production stream should read "unset -> env fallback".
#
# NOTE: v0.91.0 serializes an UNSET data_retention as 0, not null, so jq's `//` never
# fires. Test `> 0` instead — which is also exactly how the compactor decides.
curl -s -u "$O2_USER:$O2_PASS" \
  "https://bt-servant-openobserve-prod.fly.dev/api/default/streams" \
  | jq -r '.list[]
      | "\(.name)\t\(.stream_type)\t"
        + (if (.settings.data_retention // 0) > 0
           then "retention=\(.settings.data_retention)d (explicit, overrides the env fallback)"
           else "retention=unset -> env fallback" end)'

# And the machine should carry the new value
flyctl machines list --app bt-servant-openobserve-prod --json \
  | jq -r '.[].config.env.ZO_COMPACT_DATA_RETENTION_DAYS'   # expect 1825
```

Context: production telemetry went live 2026-08-05 (issue #341 / B3). Data only accumulates
from that date forward — retention is a ceiling, not a time machine.


infra/openobserve/README.md
```diff
@@ -140,18 +140,27 @@ That distinction is the whole game here, because it means:
...
+forever, while the fallback covers new streams automatically. At measured volume
+(~0.51 KB/span compressed) five years is single-digit GB, so cost was not the deciding factor.
+
+> **Editing the fallback is destructive.** Since no production stream carries its own value,
+> this variable is evaluated live by the compactor rather than stamped at write time.
+> Lowering it retroactively deletes everything older than the new value, across every
+> stream, on the next compaction — silently, with no alert and no archive.
+
+Nothing in the config prevents a stream from being set to ten years, which is how staging
+ended up there. Verify, don't assume:
```
IanLindsley, line 163: **Medium — Fix the fallback check in the read-back command.** OpenObserve v0.91.0 serializes an unset `StreamSettings.data_retention` as `0`, and jq's `//` operator does not treat `0` as missing. Consequently, every correctly unset production stream prints `retention=0`; the command can never produce the `unset -> env fallback` value that this PR tells operators to expect. Since this is described as the only proof that the single-tier policy is active, the deployment verification will falsely appear to fail. Please explicitly map non-positive values to the fallback label (for example, with an `if (.settings.data_retention // 0) > 0 then ... else ... end` expression).

IanLindsley (commented):
1 medium finding. The retention change itself is consistent with the pinned OpenObserve behavior, but the only documented post-deploy verification cannot recognize the intended fallback state. This should be treated as a request for changes; GitHub does not allow the PR author to submit a formal REQUEST_CHANGES review on their own PR.

infra/openobserve/README.md
IanLindsley: Fixed in 41f032c — good catch, and the evidence was already in front of me: an earlier live query printed `retention=0` for every stream and I read it as "unset" without noticing the documented command could never render that as such.

The read-back now tests `> 0` explicitly, which is also precisely how the compactor decides (it uses a stream's own value only when greater than zero), so the check mirrors the behaviour it verifies:

```
(if (.settings.data_retention // 0) > 0
 then "retention=<n>d (explicit, overrides the env fallback)"
 else "retention=unset -> env fallback" end)
```

Verified against the live prod instance: **46 streams, all reporting `unset -> env fallback`, zero `explicit`**. The old command reported `retention=0` for all 46. I also documented the expected output per instance so a reader knows what a pass looks like rather than inferring it.

While in there, corrected the metric-stream count from the estimated ~28 to the measured **46** — #349 added `chat_turns_total` and `auth_rejections_total` within hours, which strengthens the argument for the fallback: the set isn't static.

IanLindsley (commented):
Re-review: the repository fix for the original medium finding is correct and the jq expression now mirrors OpenObserve's `> 0` retention rule. One medium issue remains: the PR description's Verification block still contains the original broken `// "unset -> env fallback"` command, so the immediate post-merge instructions can still falsely report a correct deployment as failed. This should be treated as a request for changes; GitHub does not allow the PR author to submit a formal REQUEST_CHANGES review on their own PR.

infra/openobserve/README.md
IanLindsley: Verified in 41f032c: the README command now handles `0` correctly and mirrors the compactor `> 0` rule. Resolving this code-level thread; the still-stale duplicate in the PR description is tracked separately in the review summary.

IanLindsley (commented):
All issues addressed. The repository and PR-description verification commands now both mirror OpenObserve's `data_retention > 0` rule, the prior thread is resolved, all checks pass, and no merge gate remains. GitHub does not permit self-approval, so this clean review is submitted as COMMENT.

Issue #245, opened 2026-06-15
Connect MCP server for Open bibles to BT Servant


elsylambert:
Goal
Integrate the Open Bibles MCP server with BTServant so the bot can access and serve Bible text/translations through MCP.

Tasks
- Add the Open Bibles MCP server connection to BTServant's configuration
- Wire up the available MCP tools so BTServant can query Bible content
- Handle auth/connection setup and failure cases (server unavailable, timeouts)
- Verify BTServant can retrieve and return Bible passages via the MCP server

Acceptance criteria

- BTServant successfully connects to the Open Bibles MCP server
- Bible content can be queried and returned through the bot
- Connection errors are handled gracefully with clear logging

PR #355, opened 2026-08-10
tools: add yaapi.bible concurrency reproduction script (#245)

IanLindsley:
Adds a reproduction harness so the Open Bibles MCP server's maintainers can reproduce the HTTP 500 failures from #245 independently.

## Why

yaapi.bible failed **19 of 69** connection attempts during the #245 retest, returning an HTML `Server error (500)` page instead of a JSON-RPC response. The other four MCP servers failed zero times in the same window. Our own client was the obvious suspect, since it opens a fresh MCP session per operation rather than pooling. **That hypothesis was wrong.**

## Files

| file | role |
|---|---|
| `tools/yaapi-rca.sh` | orchestrates concurrency, collects results |
| `tools/yaapi-rca-client.mjs` | real-client arms, driven by `@modelcontextprotocol/sdk` |
| `tools/yaapi-rca-mock.py` | scriptable local MCP mock, timestamps every request |
| `tools/yaapi-rca-selftest.sh` | deterministic tests against that mock |

## Two kinds of arm

**Real-client arms (A, F, G)** run through the SDK this repo depends on. The SDK owns initialization, protocol negotiation, the post-initialize SSE stream, request ids, shutdown, and schema validation. No emulation is involved.

**Synthetic arms (B, C, D, E, H)** are plain curl POSTs. They are **not** an emulation of the SDK and no claim is made that they reproduce it. Their checks are exactly: HTTP status 200/202, body is not an HTML error page, body parses as JSON, `jsonrpc == "2.0"`, a message carrying our request id came back, no top-level error, an object result, and for calls `result.content` is a list. Deliberately shallow — enough to tell "answered" from "500'd". They exist to vary concurrency against a session while holding everything else still.

## Results (12 ops/arm)

**Real client, unmodified:**

| arm | concurrency | failures |
|---|---|---|
| F | serial | **0.0%** |
| A | 4 | **66.7%** |
| G | 4, handshake only | 25.0% |

**Synthetic sweep on one shared session, four independent runs:**

| | conc 1 | conc 2 | conc 3 | conc 4 |
|---|---|---|---|---|
| run 1 | 0.0% | 8.3% | 25.0% | 50.0% |
| run 2 | 0.0% | 8.3% | 16.7% | 41.7% |
| run 3 | 0.0% | 8.3% | 25.0% | 41.7% |
| run 4 | 0.0% | 8.3% | 25.0% | 50.0% |
| **pooled** | **0/48** | **4/48** | **11/48** | **22/48** |

- The repo's own client is clean serially and fails two thirds of the time at concurrency 4.
- Failures begin at **concurrency 2** — exactly 1 in 12 in all four runs — and rise monotonically.
- **Pooling one session does not help.** Arm E shares a session established minutes earlier and still fails 42–50%, so the client-side fix we were considering would not have solved this.

**Not established:** within the B–E sweep, concurrency and instantaneous rate co-vary, since each reused-session op is one request. The claim is "failures begin at concurrency 2 and scale with it", not "rate is irrelevant". Per-arm n = 12.

## Self-test — 9/9

`./tools/yaapi-rca-selftest.sh`, no network beyond 127.0.0.1: successful responses; malformed content (`content:[42]`) rejected by both the synthetic checks and the SDK's `CallToolResultSchema`; notification failure reported; SSE lifetime (stream opens before the tool call and closes with the client); and cleanup timing, which clusters events by the inter-arm gap and asserts per arm that no DELETE precedes that arm's last measured request.

## Response to the review

Addressed architecturally in one pass rather than by patching the three examples:

- **SDK emulation removed.** Real-client arms go through the SDK; bash does concurrency and collection only. This dissolves all three findings rather than patching them — no hand-written SSE to launch from command substitution, no hand-written DELETE in a measured window, no hand-written schema check to be incomplete.
- **No SDK-reproduction claims remain** on the synthetic arms; their validation is described exactly as implemented.
- **Cleanup** is queued and drained after the arm completes, for both kinds of arm.
- **Deterministic tests added** for all five requested behaviours.
- **PR body and script comments** rewritten to match the implementation.

Version 2.38.10 (patch — tooling only; no runtime code touched, confirmed by `git diff origin/main...HEAD -- src/` showing only the generated version string).


tools/yaapi-rca.sh
```diff
@@ -0,0 +1,195 @@
...
+run_arm() {
+  local label="$1" mode="$2" conc="$3" out="$TMP/$2.$3.out" per=$((N / $3))
+  : >"$out"
+
+  # Establish 'reuse' sessions serially first, so the control arm measures only
+  # concurrent CALLS, never concurrent session creation.
+  local -a sids=()
+  local r
+  if [[ "$mode" == "reuse" ]]; then
+    for ((w = 1; w <= conc; w++)); do
```
IanLindsley: **Medium:** This loop creates `conc` distinct sessions, so arm C does not implement the advertised "one session reused, concurrency 4"; line 165 gives each worker a different SID. B vs C therefore changes both concurrency and session count, and C cannot establish that pooling one client/session would still fail. Please either share one SID across the concurrent calls (and close it once after `wait`) or relabel the arm and narrow the pooling conclusion.

tools/yaapi-rca.sh
```diff
@@ -0,0 +1,195 @@
...
+      r=$(init_session)
+      [[ "$r" == SID* ]] || {
+        sleep 2
+        r=$(init_session)
+      }
+      sids+=("${r#SID }")
```
IanLindsley: **Medium:** If both serial initialization attempts fail, `r` is still `FAIL ...`, and this appends that non-empty string as though it were a SID. `worker` then skips its missing-SID guard, sends `mcp-session-id: FAIL ...`, and turns one setup failure into `ops` call failures. Validate `r == SID*` after the retry and record/abort the arm as an initialization failure instead of feeding the failure text to `call_tool`.

tools/yaapi-rca.sh
```diff
@@ -0,0 +1,195 @@
...
+  sid=$(grep -i '^mcp-session-id:' "$hdr" 2>/dev/null | tr -d '\r' | sed 's/^[^:]*: *//')
+  rm -f "$hdr"
+  if [[ "$body" == *"<!DOCTYPE html"* || "$body" == *"Server error"* ]]; then
+    echo "FAIL INIT_HTML500"
+  elif [[ "$code" != "200" ]]; then
+    echo "FAIL INIT_HTTP$code"
+  elif [[ -z "$sid" ]]; then
+    echo "FAIL INIT_NOSESSION"
+  else
+    echo "SID $sid"
```
IanLindsley: **Medium:** A 200 plus a session header is not the complete MCP initialization used by the real SDK. `Client.connect()` validates the negotiated `InitializeResult`, sends the required `notifications/initialized` notification, and uses the negotiated protocol for subsequent requests; this helper immediately returns the SID and omits those steps. Because the PR calls arm A "our client's real pattern," this protocol-level difference weakens the reproduction and can attribute behavior to the wrong request sequence. Complete and validate the initialization lifecycle before returning `SID`.

tools/yaapi-rca.sh
```diff
@@ -0,0 +1,195 @@
...
+    -H "Content-Type: application/json" -H "Accept: $ACCEPT" \
+    ${1:+-H "mcp-session-id: $1"} \
+    -d '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{
+         "name":"list_versions","arguments":{"language":"English"}}}')
+  code="${body##*@@}"
+  if [[ "$body" == *"<!DOCTYPE html"* || "$body" == *"Server error"* ]]; then
+    echo "CALL_HTML500"
+  elif [[ "$code" != "200" ]]; then
+    echo "CALL_HTTP$code"
+  elif [[ "$body" == *'"isError":true'* ]]; then
```
IanLindsley: **Medium:** Any HTTP 200 is reported as `OK` unless the body contains the exact compact substring `"isError":true`. Valid JSON can contain whitespace (`"isError": true`), and protocol failures can be top-level JSON-RPC `error` responses, so genuine failures are silently counted as successes. Parse the JSON/SSE payload and inspect both the top-level error and `result.isError` structurally.

IanLindsley (commented):
Review result: 0 critical, 0 high, 4 medium, and 3 low findings. The current experiment does not yet support its main session-pooling conclusion, and several failure paths can misclassify or omit failures. These findings warrant changes, but GitHub does not permit the PR author to submit a REQUEST_CHANGES review on their own PR, so this is posted as COMMENT.

tools/yaapi-rca.sh
```diff
@@ -0,0 +1,195 @@
...
+URL="${YAAPI_URL:-https://yaapi.bible/mcp}"
+N="${1:-20}"
+ACCEPT="application/json, text/event-stream"
+TMP=$(mktemp -d)
+trap 'rm -rf "$TMP"' EXIT
+
+# Opens a session. Echoes "SID <id>" on success or "FAIL <reason>" on failure.
+init_session() {
+  local hdr="$TMP/h.$RANDOM.$RANDOM" body code sid
+  body=$(curl -s -D "$hdr" -w '\n@@%{http_code}' -X POST "$URL" \
```
IanLindsley: **Low:** These `curl` calls have no connection or total timeout. A stalled third-party endpoint can leave a worker—and therefore `wait` and the whole reproduction—blocked indefinitely. Add configurable `--connect-timeout`/`--max-time` values to initialization, tool calls, and cleanup, and classify timeout exit codes in the result.

tools/yaapi-rca.sh
```diff
@@ -0,0 +1,195 @@
...
+        ;;
+    esac
+  done
+  [[ "$mode" == "reuse" ]] && close_session "$sid"
+  return 0
+}
+
+# $1=label $2=mode $3=concurrency
+run_arm() {
+  local label="$1" mode="$2" conc="$3" out="$TMP/$2.$3.out" per=$((N / $3))
```
IanLindsley: **Low:** Integer division silently drops the remainder, so concurrent arms run fewer than the requested `N` whenever it is not divisible by 4. With `N < 4`, they run zero operations and the percentage calculation divides by zero. Validate a positive multiple of the concurrency or distribute the remainder across workers.

tools/yaapi-rca.sh
```diff
@@ -0,0 +1,195 @@
...
+# 0 failures in 40 serial requests against 16 in 60 concurrent ones. The server fails
+# whenever two or more requests are in flight at once, regardless of session handling
+# and regardless of throughput — the fastest arm (C, 400 req/min) failed while the
+# slowest (D, 37.5 req/min) was clean.
+#
+# For the 'reuse' arms the sessions are established SERIALLY before timing starts, so
+# concurrent session creation is not silently folded into the control.
+#
+# Usage:
+#   ./tools/yaapi-rca.sh [OPS_PER_ARM]     # default 20, so 100 requests total
```
IanLindsley: **Low:** `20` means 100 measured operations, not 100 HTTP requests. Counting initialize, tool call, and DELETE traffic, the five default arms issue at least 210 HTTP requests (more when setup retries), and completing the MCP lifecycle adds more. Since this targets a third party's production server, please state the actual request load accurately.

IanLindsley (commented):
Re-review result: 0 critical, 0 high, 5 medium, and 1 low findings remain. Six original findings are substantively addressed, while the MCP lifecycle fidelity finding is only partial. The follow-up also introduces or exposes request-ID correlation, session-cleanup, response-validation, and conclusion-scope problems. These warrant changes, but GitHub does not permit the PR author to submit REQUEST_CHANGES on their own PR, so this review is posted as COMMENT.

tools/yaapi-rca.sh
```diff
@@ -0,0 +1,326 @@
...
+  verdict=$(classify notify)
+  [[ "$verdict" != "OK" ]] && {
+    echo "FAIL INIT_NOTIFY_$verdict"
+    return
+  }
+  echo "SID $sid"
+}
+
+call_tool() {
+  do_post '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{
```
IanLindsley: **Medium:** Every tool call still uses JSON-RPC ID `2`. Arm C now puts up to four requests with the same ID in flight on one session, whereas the real SDK increments `_requestMessageId` for each request. Duplicate active IDs can collide in server-side correlation, so arm C changes both concurrency and request identity and still does not faithfully model a pooled client. Generate IDs that are unique across all workers, and have the classifier verify the matching response ID.

tools/yaapi-rca.sh
```diff
@@ -0,0 +1,326 @@
...
+    echo "FAIL INIT_$verdict"
+    return
+  }
+  [[ -z "$sid" ]] && {
+    echo "FAIL INIT_NOSESSION"
+    return
+  }
+  do_post '{"jsonrpc":"2.0","method":"notifications/initialized"}' "$sid"
+  verdict=$(classify notify)
+  [[ "$verdict" != "OK" ]] && {
```
IanLindsley: **Medium:** Once `initialize` succeeds, this failure branch returns without deleting the newly allocated session. `INIT_NOTIFY_HTML500` is explicitly reported as occurring in the new run, so the reproduction can leak many server-side sessions and contaminate later arms with accumulated state/resource pressure. Close `sid` before returning the notification failure (while preserving the original token).

tools/yaapi-rca.sh
```diff
@@ -0,0 +1,326 @@
...
+  fi
+  printf '%s' "$DP_BODY" | python3 "$CLASSIFY" "$1" "$DP_CODE"
+}
+
+# Full MCP handshake: initialize -> validate -> notifications/initialized.
+# Echoes "SID <id>" or "FAIL <token>".
+init_session() {
+  local hdr="$TMP/h.$$.$RANDOM" verdict sid
+  do_post '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{
+            "protocolVersion":"2024-11-05","capabilities":{},
```
IanLindsley: **Medium (prior lifecycle finding remains partially open):** The installed SDK's `Client.connect()` sends its current `LATEST_PROTOCOL_VERSION` (`2025-11-25`), rejects negotiated versions outside `SUPPORTED_PROTOCOL_VERSIONS`, and calls `setProtocolVersion`, causing `mcp-protocol-version` to be sent on notification, tool-call, and DELETE requests. This script hard-codes `2024-11-05`, accepts any non-empty returned version, and never sends the negotiated-version header, so it still is not the SDK lifecycle claimed in lines 49-53. Preserve and validate the negotiated version and use it on every subsequent request.

tools/yaapi-rca.sh
```diff
@@ -0,0 +1,326 @@
...
+if 'error' in doc:
+    print('RPCERR'); sys.exit()
+result = doc.get('result')
+if kind == 'init':
+    if not isinstance(result, dict) or not result.get('protocolVersion'):
+        print('BADINIT'); sys.exit()
+if kind == 'call':
+    if isinstance(result, dict) and result.get('isError') is True:
+        print('TOOLERR'); sys.exit()
+print('OK')
```
IanLindsley, line 248: **Medium:** `call` classification still prints `OK` for any JSON object without a top-level error—even a notification or an object with no `result`. For SSE, it also takes the final `data:` line without locating the response whose ID matches the call. A local mock returning `{"jsonrpc":"2.0","method":"notifications/mock"}` for every tool call was reported as 0% failures. Parse SSE events, select the matching response ID, and require a valid `CallToolResult` before reporting success.

tools/yaapi-rca.sh
```diff
@@ -0,0 +1,326 @@
...
+#   C shared/conc4     50.0%      D fresh/serial      0.0%
+#   E initonly/conc4   33.3%      F per-worker/conc4 50.0%
+#
+# That is 0 failures in 24 serial operations against 24 in 48 concurrent ones. Every
+# concurrent arm fails at a similar rate whether it opens a session per call, shares
+# one session across all four workers, gives each worker its own, or never issues a
+# tool call at all. Throughput is not the variable: arm C sustained 360 req/min and
+# failed, arm D ran 28.8 req/min and was clean.
+#
+# Conclusion: the server fails whenever two or more requests are in flight at once.
```
IanLindsley: **Medium:** The experiment configures only serial and four-worker arms, so it cannot establish the exact threshold "whenever two or more requests"; failures might begin at three or four. It also does not control request rate, and the displayed rate is operations/min rather than HTTP requests/min (for example D uses four requests per op while C uses one). Please narrow the conclusion to the tested serial-vs-concurrency-4 result, or add concurrency-2/3 and rate-controlled arms with correctly labeled units.

tools/yaapi-rca.sh
```diff
@@ -0,0 +1,326 @@
...
+
+set -uo pipefail
+
+URL="${YAAPI_URL:-https://yaapi.bible/mcp}"
+CONNECT_TIMEOUT="${CONNECT_TIMEOUT:-10}"
+MAX_TIME="${MAX_TIME:-30}"
+N="${1:-12}"
+ACCEPT="application/json, text/event-stream"
+
+if ! [[ "$N" =~ ^[0-9]+$ ]] || ((N < 1)); then
```
IanLindsley: **Low:** Positive values below four are accepted, but the `conc 4` arms then start only `N` workers with nonzero work. With `OPS_PER_ARM=1`, the local smoke run labeled A/C/E/F as concurrency 4 even though each was serial. Require at least four operations for this fixed design or report the effective concurrency instead of a misleading label.

tools/yaapi-rca.sh
```diff
@@ -0,0 +1,421 @@
...
+# rate-controlled design would pace requests at a fixed inter-arrival time. Treat the
+# claim as "failures begin at concurrency 2 and scale with it", not as a proof that
+# rate is irrelevant.
+#
+# Sessions for arms B/C/D/E/H are established SERIALLY before timing starts, so
+# concurrent session creation is never folded into an arm meant to measure something
+# else. If setup fails twice, the arm reports SETUP_FAILED rather than feeding a bad
+# session id into measured requests.
+#
+# Protocol fidelity: this performs the lifecycle the installed SDK performs, because
```
IanLindsley: **Medium:** Arm A still is not the lifecycle the repository client actually runs. `openClient().close()` calls `client.close()`; in SDK 1.29, `StreamableHTTPClientTransport.close()` aborts the transport and does **not** send DELETE (`terminateSession()` is the separate method, and this repo never calls it). Conversely, after a 202 response to `notifications/initialized`, the SDK asynchronously starts a GET SSE stream, which this script omits. Adding DELETE to every fresh operation while omitting that potentially in-flight GET changes exactly the HTTP concurrency mix this RCA is measuring. Either run the fidelity arm through the installed SDK, emulate its GET/close behavior, or stop describing this arm and lifecycle as the client's current pattern.

IanLindsley (commented):
Re-review result at 893f8e2: 0 critical, 0 high, 2 medium, and 0 low findings remain. The six targeted fixes substantially improve the experiment, but response-schema validation remains incomplete and arm A still does not reproduce the repository client's actual Streamable HTTP lifecycle. These warrant changes, but GitHub does not permit the PR author to submit REQUEST_CHANGES on their own PR, so this review is posted as COMMENT.

tools/yaapi-rca.sh
IanLindsley: Re-review at `893f8e2`: ID matching and notification rejection are fixed, but this finding remains partially open. The validator only checks that `result.content` is a list; it does not validate the list elements as MCP `ContentBlock`s (and similarly does not fully validate the JSON-RPC envelope/InitializeResult). A local mock returning `result.content: [42]` for every tool call was still reported as 0% failures across all arms, while the installed SDK’s `CallToolResultSchema` rejects that payload. Please validate the result shape actually accepted by the SDK, or narrow the “well-formed CallToolResult” fidelity claim.

tools/yaapi-rca.sh
```diff
@@ -0,0 +1,503 @@
...
+  [[ -z "$sid" ]] && {
+    echo "FAIL INIT_NOSESSION"
+    return
+  }
+  do_post '{"jsonrpc":"2.0","method":"notifications/initialized"}' "$sid" "$pv"
+  verdict=$(classify notify 0)
+  if [[ "$verdict" != "OK" ]]; then
+    # The session exists even though the notification failed — release it rather than
+    # leaking server-side state into later arms.
+    close_session "$sid" "$pv"
```
IanLindsley: **Medium:** This still sends DELETE inside the measured window when `notifications/initialized` fails, contradicting lines 71-80 and differing from the actual `Client.connect()` failure path (`client.close()` aborts; it does not call `terminateSession()`). `INIT_NOTIFY_HTML500` was observed repeatedly, so these DELETEs are not hypothetical: they add concurrent requests to arms A/G and can affect the reported failure rates. Queue `sid pv` in `LEAKED` here and perform the harness DELETE after the arm, just like successful fresh sessions.

tools/yaapi-rca.sh
```diff
@@ -0,0 +1,503 @@
...
+    echo "FAIL INIT_NOTIFY_$verdict"
+    return
+  fi
+  local sse
+  sse=$(open_sse "$sid" "$pv")
```
IanLindsley: **Medium:** `sse=$(open_sse ...)` runs the function in a command-substitution subshell. The background `curl` still inherits that substitution's stdout FD—`-o /dev/null` is a curl option, not shell redirection—so Bash waits for the GET process to exit before assigning `sse`. On a server that keeps the SSE stream open, initialization blocks until `--max-time`, the returned PID is already dead, and no SSE GET overlaps the tool call. A local timing check took 403 ms for a 400 ms background process, versus 2 ms with shell-level stdout redirection. Redirect the process FDs at the shell level and verify it survives initialization, or return/manage the PID without command substitution.

IanLindsley (commented):
Re-review result at 3fefa33: 0 critical, 0 high, 3 medium, and 0 low findings remain. The latest patch addresses the prior examples, but the emulated lifecycle still injects DELETE on a common failure path, the SSE process does not remain concurrent when the server keeps the stream open, and the manual schemas still accept payloads rejected by SDK 1.29. These warrant changes, but GitHub does not permit the PR author to submit REQUEST_CHANGES on their own PR, so this review is posted as COMMENT.

tools/yaapi-rca.sh
IanLindsley: Re-review at `3fefa33`: `[42]` is now rejected, but the underlying SDK-schema finding remains open. The hand-written validator accepts `{"type":"resource_link","uri":"https://example.test"}` even though SDK `ResourceLinkSchema` also requires `name`; a local mock returning that block was reported as 0% failures across every tool-call arm. `InitializeResult` likewise checks `serverInfo.name` but not the required `serverInfo.version`. Please either validate the actual SDK 1.29 schemas completely or narrow the claim that responses satisfy those schemas.

IanLindsley: Resolved at bc36bdf. The fidelity arms now execute the installed MCP SDK, whose schemas reject malformed results; the remaining curl arms are explicitly documented as synthetic shallow probes rather than SDK emulation.

tools/yaapi-rca.sh
IanLindsley: Resolved at bc36bdf. The hand-written measured-path DELETE was removed and the real-client arms now use Client.close(). A separate current-line finding covers the narrower case where a failed SDK connect does not expose its allocated session to post-arm cleanup.

tools/yaapi-rca.sh
IanLindsley: Resolved at bc36bdf. The real-client arms now use StreamableHTTPClientTransport directly, and the self-test verifies that the SDK SSE GET overlaps the tool call and closes with the client.

tools/yaapi-rca-client.mjs
```diff
@@ -0,0 +1,79 @@
...
+
+for (let i = 0; i < ops; i++) {
+  const transport = new StreamableHTTPClientTransport(new URL(url));
+  const client = new Client({ name: 'yaapi-rca', version: '4.0' });
+  let phase = 'INIT';
+  try {
+    // connect() performs the whole lifecycle: initialize, protocol-version
+    // negotiation and validation, notifications/initialized, and the SSE stream.
+    await client.connect(transport);
+    if (transport.sessionId) process.stdout.write(`SESSION ${transport.sessionId}\n`);
```
IanLindsley: **Medium:** `SESSION` is emitted only after `client.connect()` resolves. If `initialize` succeeds and `notifications/initialized` returns 500, the transport has already stored the server-issued session ID, but `connect()` rejects before line 55; a direct run against the new `notifyfail` mock prints only `RESULT FAIL INIT_HTML500`, so `run_sdk_arm` never adds that session to `LEAKED` and post-arm cleanup cannot DELETE it. Repeated failing arms can therefore leak server sessions and contaminate later measurements, contrary to the cleanup guarantee. Report `transport.sessionId` once from the failure/finally path too, include `transport.protocolVersion`, preserve both fields in `run_sdk_arm`, and add a self-test assertion that the failed-init session is deleted after the arm.

IanLindsley (commented):
Re-review result at bc36bdf: 0 critical, 0 high, 1 medium, and 0 low findings remain. The SDK rewrite resolves the prior schema, DELETE-timing, and SSE-lifecycle findings, and the committed self-test passes 9/9. One cleanup gap remains when initialize succeeds but notifications/initialized fails. This warrants a change, but GitHub does not permit the PR author to submit REQUEST_CHANGES on their own PR, so this review is posted as COMMENT.

PR #357, opened 2026-08-11
feat(#356): org default language — rung 3 of the chat-time language cascade

IanLindsley:
## Summary

- Adds the org default language — the missing third rung of the chat-time language cascade (`@trigger` → per-user persisted selection → **org default**) — so language tuning authored in the Admin Portal finally reaches tokenless surfaces (web client, WhatsApp). Root cause and design were locked on unfoldingWord/bt-servant-admin-portal#255; the storage/API contract this implements is posted on #236 ([contract comment](https://github.com/unfoldingWord/bt-servant-worker/issues/236#issuecomment-5253667722)).
- `OrgLanguages.defaultLanguage?` slug reference (no KV migration — absent = current behavior; may point at an unpublished language for stage-then-publish, end users stay protected by the existing published filter at resolution time). New `GET`/`PUT /api/v1/admin/orgs/:org/languages-default` admin routes; `DELETE` of the language currently set as default returns 409. The default rung never writes per-user persistence, so default-riding users track org-level changes live.
- Telemetry: `trigger_classifier_result` gains `language_source` (`trigger` | `persisted` | `org_default` | `none`); the `language_not_found` warn's `source` field now includes `org_default` so a misconfigured default is visible in logs.

## v1 scope (deliberate)

Only the **default** language's tuning reaches users who never type `@language`; per-user choice remains the existing `@language` trigger on every surface. Reviewers/testers should validate against the org's default language. Per-user selector/tool is v2; #236 chain resolution stays blocked by admin-portal#172 and slots into this same hook later.

## Test plan

- 15 new unit tests (`tests/unit/language-default.test.ts`): cascade ordering across all rungs, empty-string drifted-KV default treated as absent, published filter for end users vs admin draft bypass, `PUT` body validation (set / clear / bad slug / nonexistent language / malformed body).
- Each new test was verified to go red against a deliberately broken implementation (cascade order swapped; existence check removed) before finalizing.
- `pnpm lint` / `check` / `architecture` / `test` (994 passed) / `build` all green locally and in the pre-commit hook.
- Post-merge staging check: `PUT` a default on the staging org, send a tokenless message, confirm `language_source: 'org_default'` + `language_document_injected: true` in telemetry.

## Version

Minor bump → 2.39.0 (new endpoints + new chat-time behavior).

Closes #356. Fixes unfoldingWord/bt-servant-admin-portal#255.


src/index.ts
```diff
@@ -1001,6 +1017,76 @@ app.delete('/api/v1/admin/orgs/:org/languages/:languageName', async (c) => {
...
+    logger.error('admin_action', error, { action: 'get_default_language', org });
+    return c.json({ error: 'Failed to read default language from storage' }, 500);
+  }
+});
+
+app.put('/api/v1/admin/orgs/:org/languages-default', async (c) => {
+  const org = c.req.param('org');
+  const logger = createRequestLogger(crypto.randomUUID());
+
+  const body = await c.req.json();
```
IanLindsley: **Medium:** Parse failures escape before this route's `try` block. A malformed JSON request (for example, a truncated portal payload) makes `c.req.json()` throw into Hono's default error handler, producing a generic 500 rather than the 400 promised by this endpoint's body-validation contract; it also bypasses the route's `admin_action` error logging. Please catch the syntax error and return a 400 (or move parsing into a dedicated validation boundary), and add a route-level test using invalid JSON so the response semantics are pinned.

IanLindsley (commented):
Review verdict: 1 medium-severity issue. The new default-language API misclassifies malformed request JSON as a server failure; please return a 400 and cover the route-level behavior before merge.

src/index.ts
IanLindsley: Fixed in f57a221. Both routes (the new `languages-default` PUT and the pre-existing `languages/:languageName` PUT, which had the identical flaw) now parse through a shared `readJsonBody` helper: malformed JSON returns a 400 with route-scoped `admin_action` warn logging instead of escaping to Hono's default 500. Route-level e2e tests added (`tests/e2e/languages-default.test.ts`) pinning the 400 semantics for both routes — verified red against the unfixed code — plus a set/read/409-guard/clear/delete round-trip.

IanLindsley (commented):
All issues addressed.

PR #361, opened 2026-08-14
fix(#360): trigger-only messages (@hindi, #dbs-coach) no longer crash the turn

### LINKED ISSUE #360: Typing @hindi to choose a language returns an error
### Problem

Using the @hindi mention/command to select a response language produces an error instead of switching the language.

### Steps to reproduce
Type @hindi to choose Hindi as the response language

IanLindsley:
## Summary

A message that is *only* routing tokens (`@hindi`, `#dbs-coach`) strips to the empty string, and the Anthropic API rejects the **entire** request when any user turn has empty content — so the user got a 502 instead of the switch they asked for.

- UserDO now sends the raw text the user typed when stripping empties the message.
- A new `## Trigger-Only Message` prompt section tells the orchestrator the switch already happened and to reply with one short confirmation. Mirrors the existing unmatched-triggers section.
- `orchestrate()` gains an empty-message backstop that logs at `error` level and substitutes a placeholder, so no future caller can cost a whole turn this way.
- `trigger_classifier_result` gains `trigger_only` so the path stays greppable.

**Scope: this PR changes no language behavior.** `@language` continues to do exactly what it did before — select which org language document is injected as `## Language Guidance` — and nothing else. An earlier revision of this branch added a "reply in this language" directive; it was reverted in be2da0f as out of scope for a crash report.

## Root cause

A **regression**, not a day-one bug. The original classifier had `strippedMessage: parsed.stripped_message || messageText` (`feb348b`), and the first deterministic rewrite kept it as `rest || messageText` (`18716ba`). The cascade rewrite `6f19b20` (#211/#212, 2026-05-11) computed the value a new way and the `|| messageText` fallback did not come along. It read like a defensive idiom; it was load-bearing, and it had no test.

Traced in staging logs — request `2d4d5ab8-a707-4c76-a340-e9101ce8e800`, 2026-08-13T14:34:26Z: `trigger_classifier_result` shows `requested_language: hindi`, `effective_language: hindi`, `unmatched_count: 0`, then 157 ms later `400 messages.10: user messages must have non-empty content`.

The selection itself **had already worked** — `applyTriggerOverrides` persists before orchestration runs — so affected users got the selection and no reply. No transcript corruption: `handleOrchestrationError` is typed `: never` and always rethrows, so `saveConversation` never ran on this path.

### Why no error handling caught it

It isn't an error-handling gap. Every catch block fired and logged correctly — that is how this was traceable to the millisecond. A 400 rejects the whole request, so there is nothing to salvage; rethrowing is correct, since swallowing it would return a silent empty reply; and 400 is deliberately absent from `RETRYABLE_STATUSES`. The missing defence was **input validation** — `validateChatBody` never asserted a non-empty `message`, and nothing re-asserted it after the classifier emptied it.

### Exposure

Live in production from **2026-05-13** (deploy `dd3eac7c`, v2.24.9 — the prior deploy `6800ad3b` is verifiably clean) through today: 93 days. Measured impact is small: over the 7 days of retained prod logs there were 1161 `trigger_classifier_result` events and exactly **one** with `message_stripped: true`. The preceding ~86 days are beyond retention and unknowable.

## Test plan

- [x] `pnpm test` — 1028 pass (30 new across 3 files)
- [x] `pnpm check`, `pnpm lint` (0 errors), `pnpm architecture`, prettier clean
- [x] **Every new test verified to fail against a deliberately broken version** — reverting the fixes turns 5 red across the three files while the controls stay green
- [x] `tests/unit/orchestrator-empty-message.test.ts` drives the real `orchestrate()` with a mocked `fetch` and asserts against the **actual captured request body** that no user turn is ever empty — regardless of which upstream component produced it, and including the populated-history shape that actually failed in prod
- [x] Reproduced on staging pre-fix (v2.39.0): `@hindi` → 502, `#dbs-coach` → 502, `@hindi who wrote the gospel of luke?` → 200 (control isolating it to token-only messages)
- [x] Verified on staging post-fix (v2.40.0, commit be2da0f): `@hindi` → 200, `#dbs-coach` → 200, `trigger_only: true` in `trigger_classifier_result`, and zero `non-empty content` errors after the deploy

### Observed behavior worth a reviewer's eye

Measured on staging after the revert:

| Sequence | Reply |
| --- | --- |
| `@hindi In one sentence, who was Ruth?` | English — identical to pre-fix |
| → plain follow-up | English |
| `@hindi` alone | **Hindi** ("भाषा हिंदी पर सेट कर दी गई है।") |
| → plain follow-up | **Hindi** |

The `@language`-with-a-question path is provably unchanged. On the trigger-only path the model tends to write its confirmation in the selected language, then that assistant turn anchors the rest of the thread. That is the model's own reading of the org's `## Language Guidance` document — the prompt gives no instruction about what language to confirm in, in either direction. Called out deliberately rather than left to be discovered: it is a real behavior difference from the 502 era (where there was no reply at all), and the decision was to not dictate language in either direction.

Also included: `fix(deps)` overriding nanoid to `>=3.3.18` for GHSA-2v37-7h3g-55p8. Unrelated to #360 — a newly-published high advisory in the dev tree (postcss ← vite ← vitest) that fails CI's full-tree audit gate on `main` identically. Fixed with a targeted `pnpm.overrides` entry, the pattern already used for postcss/undici/vite; the gate itself is untouched.

Closes #360.


IanLindsley (commented):
PR review: 0 critical, 0 high, 0 medium, 0 low findings.

The trigger-only fallback, orchestration boundary guard, prompt context, and regression coverage are consistent with the reported failure. Targeted tests, type-check, lint, and the current GitHub checks pass.

All issues addressed.

PR #363, opened 2026-08-18
feat(#354): recognize list_resources (translation-helps v2)

IanLindsley:
## Summary

- Adds a `translation-helps-v2` listing adapter for `list_resources`, so TC Helps v2 reports its catalog instead of showing a grey "no resource listing" chip in the admin portal. v2 is a separate ground-up rewrite of the server (2.0.0, served under `/v2` by reverse proxy), **not** an older deployment of the 7.x app — so v1 and v2 are configured side by side and each server resolves its own adapter from its own tool manifest.
- Checks `list_resources_for_language` **before** `list_resources`, so a server exposing both keeps v1's richer payload (v2 drops `organization` and `version`). Covered by an explicit ordering test rather than relying on statement order surviving a refactor.
- Asserts v2's declared `"N resource(s) available"` count against the parsed item count, mirroring `assertAquiferCount`. That line is the payload's only self-validating signal — discarding it would let a format drift return a short catalog as a confident success.
- Logs the previously silent `unsupported` path with the server's discovered tool names. That path was a bare early return, which is why telling "we don't recognize its listing tool" apart from "it genuinely has none" required a manual probe session.

Language tags pass through untranslated: v2 parses BCP-47 itself and answers region variants literally (`en-US` has its own much smaller catalog than `en`). That is the server's real answer, so unlike `toAquiferLanguage` there is nothing to truncate.

**Out of scope, deliberately:** yaapi stays grey — its `list_versions` returns *confidently wrong data* for an unmapped language (`es` → 30 resources, zero Spanish, no error), so it needs a fail-loud design in its own issue. FIA stays grey permanently — it is a content server with no catalog entity to enumerate (unfoldingWord/fia-mcp#3 closed `not_planned`). Staging goes from 1 green chip of 4 to 2.

## Test plan

- [x] `pnpm lint` — 0 errors. The new warn block pushed `listServerResources` past the 50-line limit, so it was extracted into `unsupportedOutcome` rather than suppressing the rule. Adds no new warnings; the one remaining in this file is pre-existing in `parseAquiferListing`.
- [x] `pnpm check`, `pnpm architecture`, `pnpm build`
- [x] `pnpm test` — 1043 tests, 59 files
- [x] **Mutation-tested every new path** — each was deliberately broken and the intended test confirmed red: dropped the count assert (count-mismatch test failed), swapped adapter ordering (ordering guard failed), removed the warn (log test failed), mapped `name` from `subject` instead of `abbreviation` (fixture test failed). All reverted.
- [x] Fixture captured verbatim from staging on 2026-08-18 — 11 resources, 9 distinct subjects, 7 mapping onto canonical slugs with `Greek New Testament` / `Hebrew Old Testament` degrading through `slugifySubject`.
- [ ] Post-deploy: confirm TC Helps reports `ok` with 11 items on staging and the portal shows two green chips — yaapi and FIA must stay grey.

Implements #354 — **deliberately not an auto-closing reference**. Elsy signs off before that issue is closed, so merging this PR must leave #354 open.

Also addresses unfoldingWord/bt-servant-admin-portal#269 and the staging half of unfoldingWord/bt-servant-admin-portal#279, with no portal change required.


IanLindsley:
## PR Review Status

### Critical
- None.

### High
- None.

### Medium
- None.

### Low
- None.

### Verdict
All issues addressed.

IanLindsley (commented):
Reviewed head `23212dd754c6d9fa7868cc99884d65b7469eafa4`.

Severity counts: critical 0, high 0, medium 0, low 0.

The v2 adapter matches the upstream `list_resources` contract, preserves v1 adapter precedence, fails loudly on payload/count drift, and adds useful unsupported-path telemetry. Focused tests, type-check, lint, and all current CI checks pass.

All issues addressed.

PR #369, opened 2026-08-24
Parse OBS-TF list_resources instead of treating it as TH v2

abelpz:
## Summary
- Staging reports `OBS Five Movements (obs-5m) failed to respond` because the worker runs the translation-helps v2 parser on every `list_resources` tool. obs-5m-mcp returns `{ resources: [{ id, type, available }] }`, not `{ available: [{ abbreviation, subject }] }`.
- Add an `obs-5m` listing adapter, selected when the server also exposes `fetch_obs_study_manual`, and parse live EN/ID fixtures.
- Unblocks Resource priorities ranking for OBS-TF so the 5M OBS coach can prefer that server instead of falling through to Translation Helps `get_obs_story` / `get_obs_notes` / `get_obs_questions` ([obs-5m-mcp#1](https://github.com/unfoldingWord/obs-5m-mcp/issues/1)).

## Test plan
- [ ] `pnpm test -- tests/unit/resource-listing.test.ts` (Node 22)
- [ ] After deploy to staging: Admin → Modes → 5 Movement OBS Translation → Resource priorities → Load `en`. `obs-5m` should be `ok`, not `translation-helps v2 listing missing "available" array`.
- [ ] Rank `obs-tf` first and confirm a new 5M OBS session calls `fetch_obs_study_manual` for Story 1 intro.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/services/mcp/resource-listing.ts
+++ b/src/services/mcp/resource-listing.ts
@@ -214,6 +214,70 @@ export function parseTranslationHelpsV2Listing(text: string, serverId: string): 
   return items;
 }
 
+// ─── obs-5m (OBS Five Movements / OBS-TF MCP) adapter ─────────────────────────
+
+/**
+ * obs-5m's `list_resources` shares a tool name with translation-helps v2 but
+ * returns `{ language, resources: [{ id, type, available, source, description }] }`
+ * — no top-level `available` array. Staging currently classifies that as
+ * `translation-helps v2 listing missing "available" array`, so OBS-TF never
+ * appears in Resource priorities and the 5M coach falls through to classic
+ * OBS TH tools (`get_obs_story` / notes / questions).
+ *
+ * Detected by the presence of `fetch_obs_study_manual` on the same server
+ * (capability-based, same pattern as the other adapters).
+ */
+const OBS_5M_TYPE_MAP: Record<string, string> = {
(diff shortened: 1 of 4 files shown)


abelpz:
Closing: this adapter is **not** required to fix the staging 5M OBS chat failure ([obs-5m-mcp#1](https://github.com/unfoldingWord/obs-5m-mcp/issues/1)).

Chat tool calling uses MCP **discovery** (`tools/list`), not the admin resource-listing parser. Staging already has `obs-5m` registered and the published `5mobs-translation` mode already names `fetch_obs_study_manual`. The coach called Translation Helps `get_obs_*` despite that — a prompt/allowlist issue, fixable from the admin portal (and already patched there with an explicit never-call block).

Resource priorities is the zero-worker path: it splices markdown into Tool Guidance. The listing parser only affects whether `obs-5m` rows appear in that ranking UI. Production does not register OBS-TF MCP yet, so this does not belong on `main` for the current fix.

Note: `Deploy to Staging` already ran on this PR and pushed this branch to the shared staging worker. If staging should stay on `main`, redeploy `main` to staging after this close.

abelpz:
Not needed for the staging chat fix — admin portal prompt is the right surface.

PR #370, opened 2026-08-26
feat(#333): enable prompt caching and report token spend

### LINKED ISSUE #333: Implement prompt caching for Anthropic API calls
## Scope

- [ ] Identify which code paths send requests to the Anthropic API and which of
      them reuse a large, stable prefix (system prompt, tool definitions,
      reference documents)
- [ ] Estimate current spend on those paths and projected savings from caching
- [ ] Add `cache_control` breakpoints to the qualifying requests
- [ ] Verify cache hits in the API response usage fields
      (`cache_creation_input_tokens` / `cache_read_input_tokens`)
- [ ] Measure actual cost and latency change against the pre-caching baseline

IanLindsley:
Closes #333.

## The problem

Anthropic has been emailing unfoldingWord that our prompt cache hit rate is low. It is not low — it is **zero**. `buildMessageBody()` sent `system` as a plain string and no `cache_control` existed anywhere in the repo. Cache writes happen only at a breakpoint, and the backward lookback finds *prior writes*, not "stable-looking content" — so there was nothing to write and nothing to read.

`src/services/claude/orchestrator.ts` is the only Anthropic call site in the repo, and no sibling BT Servant service calls the API in production. One function held the entire opportunity.

Measured on production 2026-08-26, re-sent at full input price on **every** call, on **every** one of up to 100 orchestration iterations per turn:

| Active mode | stable prefix (tools + catalog + prompt slots) |
|---|---:|
| _(none — org defaults)_ | ~7,950 tok |
| `dbs-coach` | ~4,900 tok |
| `translation-coach` | ~9,200 tok |
| `cbbt-mentoring` | ~11,200 tok |
| `spoken-mode` | ~17,300 tok |

Cache reads cost 10% of base input, so we were paying roughly 10x for content that never changes.

## What this does

**1. Splits the system prompt — without reordering it.** `buildSystemPromptBlocks()` returns `{ stable, volatile }`; `buildSystemPrompt()` keeps its signature and is now `stable + volatile`, so every pre-existing test asserts on exactly the string it did before. The cut lands on a seam that already existed in the assembly order (after `instructions`). The `'\n\n'` separator is carried *inside* the split rather than re-joined by the caller — otherwise a lost separator would pass an equality test that re-adds it.

**2. Two cache breakpoints**, 2 of the 4 available slots:
- explicit, on the stable system block — render order is `tools` → `system` → `messages`, so one entry covers the tool definitions *and* the prompt prefix;
- top-level `cache_control` (automatic caching) for the orchestration loop, where each iteration re-sends every tool result accumulated so far. The breakpoint advances every iteration, keeping consecutive writes well inside the 20-block lookback window.

The prefix is keyed by (org, mode) and shared across every user, client and conversation in that scope. `clientId`, the memory TOC and the speaker name stay in the volatile block — that is what avoids the multi-tenant zero-hit-rate trap.

**3. Token telemetry that was missing entirely.** `response.usage` was parsed and discarded, so we could not state what we spent and therefore could not state what we saved. `claude_response` now carries input / output / cache-write / cache-read plus a derived billable-equivalent total, and OTel counters are emitted labelled by `model` and `mode`.

**4. Cache-miss attribution.** An FNV-1a fingerprint of the stable block on `claude_request`, so a future hit-rate dip can be traced to a shrunken MCP catalog versus an edited prompt override. Both change the prefix; both are otherwise invisible.

## Reporting savings needs no baseline

Caching does not change the prompt, so the counterfactual is exact rather than estimated:

```
savings_tokens = (input + cache_write + cache_read) − billable_equivalent
               = 0.9×read − 0.25×write_5m − 1.0×write_1h
savings_usd    = savings_tokens × $3.00/MTok   (claude-sonnet-4-6)
```

The multipliers encoded in the worker are model-independent **pricing ratios, not prices**, so no price table sits in the request path. Full recipe in `docs/plans/prompt-caching.md`.

## How safe is this?

The savings come from *marking* text as cacheable, not from moving it. Three independent guards:

- a **byte-identity test** (`stable + volatile === buildSystemPrompt(...)`) across a 13-case option matrix;
- the **1,081 pre-existing tests** pass unchanged, including the suites that assert on the joined prompt's content and ordering;
- Anthropic's own guarantee: *"Prompt caching has no effect on output token generation. The response you receive is identical to what you would get if prompt caching were not used."*

**Honest caveat:** `system` does change from a string to a two-element array of text blocks. Whether the API inserts anything between adjacent system text blocks is not documented, so the strict claim is "same text, same order", not "provably identical tokenization". Staging validation below closes that gap.

## Verification

`pnpm lint` / `check` / `test` / `architecture` / `build` all green — **1,081 tests pass**, 38 of them new.

**Every new test was deliberately broken and confirmed red before being trusted**, per CLAUDE.md — green-on-arrival proves nothing:

| Mutation | Result |
|---|---|
| client section leaked into the stable block | 2 failed |
| separator dropped from the volatile block | 1 failed |
| breakpoint moved to the volatile block | 1 failed |
| top-level rolling breakpoint removed | 1 failed |
| write multiplier corrupted (1.25 → 1.0) | 4 failed |
| usage field dropped from the log | 2 failed |
| stable-prefix hash removed | 1 failed |
| seam moved one section later | 2 failed |

Worth noting: the byte-identity test *alone* cannot catch a lost separator, because both sides of that assertion come from the same function. That is why the separator and seam-boundary assertions exist separately — the mutation run is what surfaced it.

**Staging acceptance signal (not yet run):** two chats within 60s as a synthetic user, then read `claude_response` via `cf-logs`. `cache_read_input_tokens` must be **0 on the very first iteration and > 0 on every subsequent iteration and on the second chat**.

## Deliberately out of scope

- The **~1,400 tokens of stable text stranded behind volatile sections** (`client_instructions`, `memory_instructions`, `AUDIO_GUIDANCE`, `MEDIA_FORMATTING_RULES`, `closing`). Recovering them means physically moving instruction text — the only behavior-affecting change in this effort. Worth ~$50/month against a real chance of shifting how the model weighs those sections; needs its own PR and a staging A/B.
- **1h TTL** — ship on the 5-minute default, revisit once the new metrics show a real cross-turn hit ratio.
- **Cross-repo reporting** in `bt-servant-telemetry` (D1 columns, Zulip digest line, dashboard KPI).

Version bumped to 2.42.0 (minor — new behavior + new telemetry).


IanLindsley (commented):
Review complete at head 11f0dd2. Critical: 0; High: 0; Medium: 0; Low: 0. The cache boundaries, Anthropic request shape, streaming/non-streaming usage accounting, telemetry labels, and regression coverage are consistent with the intended behavior. All issues addressed.

